PR #4, opened 2018-03-07
Added csurf middleware

tchow8:
[DIV-1056](https://tools.hmcts.net/jira/browse/DIV-1056)

#### What this change does?

Adds in the csurf middleware and a hidden input field taking in a csrfToken value to every form and client side POST. This will prevent cross-site scripting attacks.

For more understanding, see the below link:
https://github.com/pillarjs/understanding-csrf

There should be no visible effect on the user. Other sites such as phising sites will instead return an error page when they attempt to make a request to our site without a valid csrfToken.

Note that all requests towards the node backend must have a valid csrfToken as part of the body, query etc. otherwise a 403 will be returned.

#### Why make this change?

This change is a security requirement to prevent Cross-Site Scripting attacks.

#### How is the change implemented?

Added the npm module 'csurf' which is an express middleware that generates a CSRF secret for that session, which is used to validate generated CSRF tokens.
Validation occurs on all non-GET requests as thse requests can introduce session changes, which must be protected.
A hidden input is placed in every form, and populated with the csrfToken. Without this form the validation on every form request will fail as the csrfToken could not be validated. Since this input is populated by the node backend, external sites attempting to utilise cross site requests will not have access to a valid token.

For the Upload and Download javascripts, and the multipart form upload when javascript is disabled, the csrfToken is extracted from the page itself (which will only have a valid csrfToken if it is an actual page from us) and then appended as query parameters on the requests.

Unit tests utilising requests have been modified to add a valid csrfToken on request. For the a11y tests, the csurf module is stubbed out so pages can be retrieved correctly.

An additional End to End test checks that the csrfToken is properly populated on the first question page, then another test checks that the application does not continue when the csrfToken is invalid.


app.js
```diff
@@ -125,6 +126,24 @@ exports.init = () => {
 
   app.use(middleware.locals);
 
+  app.use((req, res, next) => {
+    next();
```
dtaylor7: is this necessary?

app.js
```diff
@@ -125,6 +126,24 @@ exports.init = () => {
...
+
+  app.use(csurf(), (req, res, next) => {
+    res.locals.csrfToken = req.csrfToken();
+    next();
+  });
+
+  app.use((error, req, res, next) => {
+    if (error.code === 'EBADCSRFTOKEN') {
+      res.status(statusCode.FORBIDDEN);
+      res.send('Invalid CSRF token');
```
dtaylor7: redirect to generic error page?

app.js
tchow8: Ah, nice catch. I think I was using this for some debugging but forgot to remove it.

app.js
tchow8: Done. Is there a way to log that it was a csrf error here though? Otherwise it might be hard to tell why we are redirecting to a generic-error page.

tchow8:
Holding on merging before I test. I will find time to deploy this branch onto DevA and verify everything still works.

tchow8:
Ready to merge. Deployed a version from the PR branch into DevA for testing purposes. Full journey seems to be working:
Petitioner Questions
Upload Files
Delete Files
Save and Close
Resume
Delete Application
Submit
Payment
---
Invalid CSRF redirects to /generic-error page

PR #5, opened 2018-03-08
Retrieve divorce application fee from Fee Register

dtaylor7:
[DIV-1976 - Retrieve divorce application fee from Fee Register](https://tools.hmcts.net/jira/browse/DIV-1976)

#### What this change does?
 - Adds service to retrieve fee for divorce
 - Adds middleware to enable any page that requires fee to retrieve it & cache in redis
 - Adds new middleware to all pages requiring divorce fee
 - Updates new fee config

### DIFF
--- a/app/middleware/updateApplicationFeeMiddleware.js
+++ b/app/middleware/updateApplicationFeeMiddleware.js
@@ -3,11 +3,19 @@ const feeRegisterService = require('app/services/feeRegisterService');
 const mockFeeReigsterService = require('app/services/mocks/feeRegisterService');
 const logger = require('@hmcts/nodejs-logging').getLogger(__filename);
 const ioRedis = require('ioredis');
+const ioRedisMock = require('app/services/mocks/ioRedis');
 
 const redisHost = CONF.services.redis.host;
 
 // redisClient is a let so it can be rewired in tests
-let redisClient = new ioRedis(redisHost); // eslint-disable-line prefer-const
+let redisClient = {};
+
+if (process.env.NODE_ENV === 'testing') {
+  redisClient = ioRedisMock();
+} else {
+  redisClient = new ioRedis(redisHost); // eslint-disable-line prefer-const
+}
+
 redisClient.on('error', error => {
   logger.error(error);
 });
--- a/app/middleware/updateApplicationFeeMiddleware.js
+++ b/app/middleware/updateApplicationFeeMiddleware.js
@@ -1,0 +1,46 @@
+const CONF = require('config');
+const feeRegisterService = require('app/services/feeRegisterService');
+const mockFeeReigsterService = require('app/services/mocks/feeRegisterService');
+const logger = require('@hmcts/nodejs-logging').getLogger(__filename);
+const ioRedis = require('ioredis');
+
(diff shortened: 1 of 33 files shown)


app/services/submission.js
```diff
@@ -54,7 +54,7 @@ const generatePaymentEventData = (session, response) => {
       PaymentDate: formatDate(dateCreated),
       PaymentAmount: amount,
       PaymentStatus: paymentStatus,
-      PaymentFeeId: CONF.commonProps.applicationFeeCode,
```
lukaszdrynski: Can you also remove this value from the default configuration file?

app/middleware/updateApplicationFeeMiddleware.js
```diff
@@ -0,0 +1,46 @@
...
+const applicationFeeQueryParams = 'service=devorce&jurisdiction1=family&jurisdiction2=family%20court&channel=default&event=issue';
+
+const getFeeFromService = () => {
+  const service = CONF.deployment_env === 'dev' ? mockFeeReigsterService : feeRegisterService;
+  const options = { queryParameters: applicationFeeQueryParams };
+
+  return service.get(options)
+    .then(response => {
+      // set fee returned from fee register to global CONF
+      CONF.commonProps.applicationFee = response;
```
lukaszdrynski: can you you use `CONF.commonProps.applicationFee.fee_amount` instead (here and the rest of the file)?

app/middleware/updateApplicationFeeMiddleware.test.js
```diff
@@ -0,0 +1,58 @@
...
+      underTest.__set__('redisClient', ioRedisClient);
+      req = { session: {} };
+      res = { redirect: sinon.stub() };
+    });
+    it('gets application fee from fee register', done => {
+      underTest.updateApplicationFeeMiddleware(req, res, () => {
+        expect(ioRedisClient.get.calledOnce).to.eql(true);
+        expect(ioRedisClient.set.calledOnce).to.eql(true);
+        expect(ioRedisClient.expire.calledOnce).to.eql(true);
+        expect(CONF.commonProps.applicationFee)
```
lukaszdrynski: can you you use `CONF.commonProps.applicationFee.fee_amount` instead (here and the rest of the file)?

dtaylor7: no, its an object now.... so testing to make sure object is correct

app/middleware/updateApplicationFeeMiddleware.js
dtaylor7: no, its an object now.... need all info so that we have code etc for payments

app/services/submission.js
dtaylor7: how come?

lukaszdrynski: In default.yaml you have 
```commonProps:
 applicationFee: 
   code: 'X0165'
 applicationFeeCode: 'X0165'
```
Do you need to keep them both? you just use `applicationFee.code` and remove the other one

dtaylor7: ahh yea good find, no i don't...removed `applicationFeeCode` from config.

PR #6, opened 2018-03-08
Div 2000 logout

niyioyelade:
https://tools.hmcts.net/jira/browse/DIV-2000

Add Login functionality to the divorce application on save&close, exit and done.

This is the frontend bit of the code. It requires the corresponding idam express middleware to be built and added to node modules for this to work..

DestroySessionStep has been updated to make a middleware call to idam to delete session token.


niyioyelade:
Version number increase for middleware dependency omitted.

MOsinloye:
This has been tested and passed.

Ready for merge

PR #7, opened 2018-03-08
Div 2000 logout

niyioyelade:
https://tools.hmcts.net/jira/browse/DIV-2000

Add Login functionality to the divorce application on save&close, exit and done.

This is the frontend bit of the code. It requires the corresponding idam express middleware to be built and added to node modules for this to work..

DestroySessionStep has been updated to make a middleware call to idam to delete session token.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   ],
   "dependencies": {
     "@hmcts/div-feature-toggle-client": "https://github.com/hmcts/div-feature-toggle-client#1.0.7",
-    "@hmcts/div-idam-express-middleware": "https://github.com/hmcts/div-idam-express-middleware#3.0.5",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
   ],
   "dependencies": {
     "@hmcts/div-feature-toggle-client": "https://github.com/hmcts/div-feature-toggle-client#1.0.7",
-    "@hmcts/div-idam-express-middleware": "https://github.com/hmcts/div-idam-express-middleware#3.0.4",
+    "@hmcts/div-idam-express-middleware": "https://github.com/hmcts/div-idam-express-middleware#3.0.5",
```
lukaszdrynski: You also need to upgrade yarn file

PR #17, opened 2018-03-14
Div 1956 invalid case ref hitting gov pay

niyioyelade:
Throwing an error when Case ID is not provided.

https://tools.hmcts.net/jira/browse/DIV-1956

Promise rejects when case ID/reference is not provided
Error is logged in this scenario as well
Unit tests updated to cover this scenario
This will prevent payments with invalid case ID being forwarded to prod Gov Pay subsequently

### DIFF
--- a/app/steps/pay/pay-online-only/index.js
+++ b/app/steps/pay/pay-online-only/index.js
@@ -76,6 +76,7 @@ module.exports = class PayOnline extends Step {
     const siteId = get(req.session, `court.${req.session.courts}.siteId`);
 
     if (!caseId) {
+      // console.log('Case Reference is missing');
       logger.error('Case Reference is missing');
       res.redirect('/generic-error');
     }
--- a/app/steps/pay/pay-online-only/index.js
+++ b/app/steps/pay/pay-online-only/index.js
@@ -86,7 +86,7 @@ module.exports = class PayOnline extends Step {
 
     if (!caseId) {
       logger.error('Case ID is missing');
-      res.redirect('/generic-error');
+      return res.redirect('/generic-error');
     }
 
     // Initialise services.
--- a/app/steps/pay/pay-online-only/index.js
+++ b/app/steps/pay/pay-online-only/index.js
@@ -76,8 +76,7 @@ module.exports = class PayOnline extends Step {
     const siteId = get(req.session, `court.${req.session.courts}.siteId`);
 
     if (!caseId) {
-      // console.log('Case Reference is missing');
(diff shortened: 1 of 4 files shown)


app/steps/pay/pay-online-only/index.test.js
```diff
@@ -188,7 +204,7 @@ describe(modulePath, () => {
         it('redirects to the gov.uk payment page', done => {
           // Act.
           const featureMock = featureTogglesMock
-            .when('onlineSubmission', true, testCustom, agent, underTest, cookies, response => {
+            .when('onlineSubmission', true, testCustom, agent, underTest, [], response => {
```
Rob-Doherty: Is this a test that not including cookies should still redirect to payments page? Is this definitely correct?

app/steps/pay/pay-online-only/index.js
```diff
@@ -84,6 +84,11 @@ module.exports = class PayOnline extends Step {
     const caseId = req.session.caseId;
     const siteId = get(req.session, `court.${req.session.courts}.siteId`);
 
+    if (!caseId) {
+      logger.error('Case ID is missing');
+      res.redirect('/generic-error');
```
dtaylor7: needs to be `return res.redirect('/generic-error');`

so as to not execute the rest of the code in this function

niyioyelade: return statement added. Thanks

app/steps/pay/pay-online-only/index.test.js
niyioyelade: This is now sorted. Cookies has been passed in. Thanks.

app/steps/pay/pay-online-only/index.test.js
```diff
@@ -107,34 +107,61 @@ describe(modulePath, () => {
...
+
+        withSession(done, agent, session);
+      });
+      it('redirects to the generic error page', done => {
+        // Act.
+        const featureMock = featureTogglesMock
+          .when('onlineSubmission', true, testCustom, agent, underTest, cookies, response => {
+            // Assert.
+            expect(response.status).to.equal(statusCodes.MOVED_TEMPORARILY);
+            expect(response.header.location).to.equal('/generic-error');
```
dtaylor7: would add test here to ensure that rest of code isn't ran... i.e. `expect(serviceTokenService.setup.called).to.eql(false);`

niyioyelade: Good spot. Done.

PR #25, opened 2018-03-20
updated already submitted url

tchow8:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

When the user attempts to start an application (based on their IDAM account) but they already have a submitted application, then it redirects them to the /error-application-submitted page.
This is not user friendly as it implies there is an error, so the url has been requested to be changed to /application-submitted.

## Type of change

# How Has This Been Tested?

Existing tests have been modified to match this change

**Test Configuration**:

* Hardware:
* O/S and version:
* JDK:

# Checklist:


app/steps/error-submitted/index.js
```diff
@@ -3,7 +3,7 @@ const runStepHandler = require('app/core/handler/runStepHandler');
 
 module.exports = class SubmittedError extends Page {
```
Rob-Doherty: Is this still an `Error` class? If not should the class and folder name change as well?

tchow8: Not an Error class, just named as such it seems. I'll rename if you want.

Rob-Doherty: that would be good for clarification, otherwise it could get confusing in a future refactor

tchow8:
Moved to PR #27

PR #31, opened 2018-03-22
fix sitemap

dtaylor7:
[DIV-1101 - fix sitemap](https://tools.hmcts.net/jira/browse/DIV-1101)


app/core/steps/Step.test.js
```diff
@@ -373,11 +362,11 @@ describe(modulePath, () => {
     it('memoises the router', () => {
       expect(step.router).to.eql(step.router);
     });
-    it('throws an error if url is not implemented', () => {
+    it('returns null if url is not implemented', () => {
```
Rob-Doherty: will this `null` throw an error upstream? and if not, do we want it to?

app/steps/pay/card-payment-status/index.js
```diff
@@ -128,6 +128,15 @@ module.exports = class CardPaymentStatus extends Step {
     return '/pay/card-payment-status';
   }
 
+  get nextStep() {
+    return {
```
Rob-Doherty: was this code originally missed?

Rob-Doherty: can you add unit test coverage for it just so it's fully covered, especially given it's a payment change

app/middleware/draftPetitionStoreMiddleware.js
```diff
@@ -134,15 +135,13 @@ const saveSessionToDraftStoreAndReply = function(req, res, next) {
     return client
       .saveToDraftStore(authToken, session)
       .then(() => {
-        const successStatusCode = 200;
         res
-          .status(successStatusCode)
+          .status(statusCode.OK)
```
Rob-Doherty: nice cleanup!

app/core/steps/Step.test.js
mattyod: No and we don't want it to. It should be covered by tests. What you don't want is checks such as `typeof` and `instanceof` throwing uncaught errors. Which is what was happening (in the case of nextStep) for sitemap view.

app/steps/pay/card-payment-status/index.js
mattyod: This change was made before I picked this branch up:
https://github.com/hmcts/div-petitioner-frontend/pull/31/commits/e1f626141acab254758b692a9a1057ba26a2dd6e

But appears to have tests covering the outcomes.
Here:
https://github.com/hmcts/div-petitioner-frontend/blob/fix-sitemap/app/steps/pay/card-payment-status/index.test.js#L151
And Here:
https://github.com/hmcts/div-petitioner-frontend/blob/fix-sitemap/app/steps/pay/card-payment-status/index.test.js#L198

dtaylor7: yes, this is required ( although not used ) by the site map so it knows what page comes next...

PR #34, opened 2018-03-26
Div-2272 - change links when browsing on a mobile device

richard-diamond:
# Description

Fix of ticket DIV-2272 (first issue). The fix involved creating a word-break stylesheet and adding that stylesheet to the change link column of each segment of the Check Your Answers page.

Fixes # (issue)

DIV-2272

## Type of change

# How Has This Been Tested?

Problem can be reproduced by reducing the width of the browser window accessing the Check Your Answers page. At a certain point, the "Change" link text splits over two lines. After the change, The "Change" link text no longer splits in two

# Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [x ] My changes generate no new warnings


richard-diamond:
That's a good point, but I also tested it with various browsers on my computer, and the same problem occurs when you reduce the width of the browser window. From that I deduced that the problem was cross-platform, but only showed up because of the inherently narrow browser space on a mobile phone screen. It's for that reason I implemented this solution to work on all platforms.

dnenkov:
I think the `@media` tag captures the browser window width. 

Also, instead of setting a class to each third `td` you can use this selector https://www.w3schools.com/cssref/sel_nth-of-type.asp

richard-diamond:
requested changes implemented as requested

PR #52, opened 2018-04-03
Disable the rate limiter

dnenkov:
# Description

Disable the rate limiter as it puts requests from different users into the same key in Redis. This can potentially block real users once the load increases. 

Proper solution will be to pass a header from the load balancer which will allow us to split the requests by users. To do that, however, we will need changes to Nginx/F5 on tactical and given that we are migrating to CNP soon, it is not worth investing the time now. 

Once we migrate to CNP we will have to re-enable the rate limiter. [DIV-2402](https://tools.hmcts.net/jira/browse/DIV-2402) was created to tackle this.

Fixes # [DIV-2145](https://tools.hmcts.net/jira/browse/DIV-2145)

## Type of change

# How Has This Been Tested?

Tested locally

# Checklist:


app.js
```diff
@@ -122,7 +122,7 @@ exports.init = () => {
 
   app.set('trust proxy', 1);
   app.use(sessions.prod());
-  if (process.env.NODE_ENV === 'production') {
+  if (process.env.NODE_ENV === 'production' && CONF.rateLimiter.enabled) {
```
dtaylor7: might as well remove production test then

PR #54, opened 2018-04-04
Fix focus issue on document upload, add browser detection functionality

borislavpetrovhmcts:
# Description

[DIV-2278 - DIV-Accessibility- Java Script- petitioner-respondent/address pages (A)](https://tools.hmcts.net/jira/browse/DIV-2278)

Removes custom tab order and prevent double trigger of upload modal

## Type of change

# How Has This Been Tested?

Manually tested.

# Checklist:

### DIFF
--- a/app/assets/javascripts/documentUpload.js
+++ b/app/assets/javascripts/documentUpload.js
@@ -172,6 +172,8 @@
           if([13, 32].includes(e.keyCode)) { // pressed RETURN or SPACE
             e.preventDefault();
             self.$zone.trigger('click');
+            e.preventDefault();
+            return false;
           }
         });
       },
--- a/app/assets/javascripts/documentUpload.js
+++ b/app/assets/javascripts/documentUpload.js
@@ -171,8 +171,8 @@
         $(document).on('keydown', 'a.faux-link, .dz-clickable', function(e) {
           if([13, 32].includes(e.keyCode)) { // pressed RETURN or SPACE
             e.preventDefault();
+            e.stopPropagation();
             self.$zone.trigger('click');
-            e.preventDefault();
             return false;
           }
         });
--- a/app/steps/marriage/upload/template.html
+++ b/app/steps/marriage/upload/template.html
@@ -83,7 +83,7 @@
 
             <p>{{ content.limitations }}</p>
 
-            <div class="document-upload" tabindex="0">
+            <div class="document-upload">
               <div class="fallback">
                 <form method="POST" accept-charset="utf-8" class="form" id="form"
                   enctype="multipart/form-data"
@@ -105,7 +105,7 @@
                     <p class="dz-message-copy">
                       {{ content.dragAndDrop }}<br>
                       <strong>{{ content.or }}</strong><br>
-                      <a href="#" class="faux-link" tabindex="-1">{{ content.clickToChooseFile }}</a>
+                      <a href="#" class="faux-link">{{ content.clickToChooseFile }}</a>
                     </p>
                   </div>
                 </div>


app/core/handler/runStepHandler.test.js
```diff
@@ -34,7 +35,8 @@ describe(modulePath, () => {
         return '/test';
       },
       method: 'POST',
-      sessionID: 'qwertyuiop'
+      sessionID: 'qwertyuiop',
+      headers: { 'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.162 Safari/537.36' }
```
dnenkov: I think we should fine a way to make the tests work without providing the `user-agent` header. Otherwise we are going to discover many other tests that need this.

app/steps/marriage/upload/template.html
```diff
@@ -105,7 +108,13 @@ <h1 class="heading-large">
                     <p class="dz-message-copy">
                       {{ content.dragAndDrop }}<br>
                       <strong>{{ content.or }}</strong><br>
-                      <a href="#" class="faux-link" tabindex="-1">{{ content.clickToChooseFile }}</a>
+                      <a href="#" class="faux-link" tabindex="
```
dnenkov: It will be good to write tests that verify that the `tabindex` is correct for IE and for other browsers

app/core/handler/runStepHandler.test.js
borislavpetrovhmcts: Well, This is a mock. So wherever the template rendering is tested the mock will be required. As browser detection is now part of runstephandler for the rendering of nunjucks templates.

app/steps/marriage/upload/template.html
borislavpetrovhmcts: Tabindex is hardcoded, as required by the story, for IE is 0, for everything else is -1. It does not come from a content variable, so it will be hard to test... I am open for suggestions how to test it though

app/core/handler/runStepHandler.test.js
dnenkov: Just looked at the code again and I think that we should be more careful when we trust that the browser will send the `user-agent` header. If the header is not sent, the code will throw an exception which we are not catching. 

I think there are quite a lot of scenarios where this header might not be sent, for example: 
* VPN or router configuration at the client 
* Load balancer configuration on our side

I think something like this is going to help 

```javascript
const browserName = (req.headers && req.headers['user-agent']) ? browser(req.headers['user-agent']).name : null;
```

dnenkov:
I actually agree with @dtaylor7 that we should do this in the client.

dtaylor7:
this should get you started: https://www.quirksmode.org/css/condcom.html

borislavpetrovhmcts:
@dtaylor7 Support for conditional comments has been removed in Internet Explorer 10 standards and quirks modes for improved interoperability and compliance with HTML5.

dtaylor7:
yea but i assumed this tabIndex issue was for older IE versions?

borislavpetrovhmcts:
@dnenkov and @dtaylor7  How are we going to this in the client if the javascript on the client side is disabled ?

dtaylor7:
if not then you will have to use javascript

borislavpetrovhmcts:
@dtaylor7  Tabindex issue is on all IE (including 11), I have tested it on real windows VMs

dnenkov:
@borislavpetrovhmcts I think we should use javascript. Regarding the users with javascript disabled, I think we are safe to say that if the user uses a screen reader, IE and has disabled their javascript - they will experience this issue. We have made similar decision for other bugs as well

borislavpetrovhmcts:
@dnenkov We have to make sure our app is functional with client side javascript disabled. If you are not able to upload files with your javascript off - the app will not be functional.

dtaylor7:
its ok, there is already a basic form fallback when javascript is disabled

borislavpetrovhmcts:
Even with the basic form, when using keyboard only you have issues with IE, because of the markup the tab order is not right

dtaylor7:
ok that one will need to be fixed too then i guess

app/assets/javascripts/documentUpload.js
```diff
@@ -172,6 +172,8 @@
           if([13, 32].includes(e.keyCode)) { // pressed RETURN or SPACE
             e.preventDefault();
             self.$zone.trigger('click');
+            e.preventDefault();
```
dnenkov: do you need both `e.preventDefault();`

dtaylor7: no, good spot have replaced on with `e.stopPropagation();`

PR #63, opened 2018-04-10
change dash to unicode value

dtaylor7:
# Description
[DIV-2412 - DIV- Accessibility -Case Number Format- page (AAA)](https://tools.hmcts.net/jira/browse/DIV-2412)

Change the - to a unicode value so the screen reader will pick up its correct meaning

## Type of change

# How Has This Been Tested?

Manually using a screen reader

# Checklist:

### DIFF
--- a/app/filters/nunjucks.js
+++ b/app/filters/nunjucks.js
@@ -2,7 +2,7 @@ const newFilterInsertHyphen = input => {
   if (!input || (typeof input !== 'number' && typeof input !== 'string')) {
     return input;
   }
-  return `${input}`.match(/.{1,4}/g).join(' - ');
+  return `${input}`.match(/.{1,4}/g).join(' &#8208; ');
 };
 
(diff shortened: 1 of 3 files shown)


nsingamsetty:
Test Passed using Windows Laptop - NVDA on Chrome browser.  - https://pr-63-div-pfe-preview.service.core-compute-preview.internal/  - Please refer to comments updated for DIV-2412 jira for more details.
The voice over does not state 'to', but rather just the numbers (it does not state 'hyphen'). 
Mac VoiceOver, & NVDA doesn’t inform users about dashes. JAWS should inform users about dashes.
Accessibility team assume that what we currently have should be fine with the existing behaviour
Also there is some text appearing like(&#8208) in between the reference number on Mac chrome version 66. I have tested this and it is working as expected hence merging the PR - Tested here https://pr-63-div-pfe-preview.service.core-compute-preview.internal/index. PR Build also successful.

PR #68, opened 2018-04-13
Cnp

tchow8:
# Description

This PR contains all the changes applied to the CNP branch in order to deploy into AAT and Demo on CNP.

The main changes are:
Adding in a Jenkinsfile_CNP (for AAT) and Jenkinsfile_parameterized (for Sandbox) which will get picked up by the relevent Jenkins build agent for deployment.

Adding in terraform var files which define the deployment parameters and application environment variables for Sandbox, AAT and Demo.

As EM Client API should now point to CNP, which should implement PaloAlto, ClamAV has been removed as part of this PR, and tests adjusted. The token to be passed to EM Client API has been changed from the 'authorizationToken' header to a 'Authorization' header as per standard now.
As CNP urls return SSL cert errors for self signed certificates, where we are using our own Divorce Backend Services, a function to replace the https:// with http:// is in the healthcheck.js. Http:// is still a valid protocol, which should be acceptable for internal connections.

Numerous package.json script changes were made to be compatible with CNP deployment script, as such please make note of what may have changed. For example, 'yarn test' now only runs unit tests and not coverage, a11y and e2e. Those have been moved to 'yarn test-all'.

Functional and Smoke Tests have been implemented and working locally (as CNP does not have the xfvb plugin to run headless browser testing), but due to a docker-compose version being outdated issue on CNP, these tests are currently disabled and simply output an echo.

Logging has been updated to use the latest HMCTS nodeJs logger with support for AppInsights. As such the code has very slightly changed to use a Logger object instead of directly using the module.export. This is an extremely minor change from' logger.getLogger' to 'logger.Logger.getLogger' for example.

Fixes # (issue)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

CNP changes are mostly non-functional. No new features has been added, but old features have been slightly modified. Testing was done locally and a deployment to Demo CNP was done using the same code base.

**Test Configuration**:

* Hardware: Mid-range Laptop
* O/S and version: Ubuntu 17
* JDK: Java 8

# Checklist:


infrastructure/demo.tfvars
```diff
@@ -0,0 +1,27 @@
+env = "demo"
+deployment_env = "preprod"
+
+vault_section = "preprod"
+
+packages_environment = "preprod"
+packages_version = "2.2.0"
```
dnenkov: Version here as well

infrastructure/main.tf
```diff
@@ -0,0 +1,213 @@
...
+  source = "git@github.com:hmcts/moj-module-webapp.git?ref=master"
+  product = "${var.product}-${var.microservice}"
+  location = "${var.location}"
+  env = "${var.env}"
+  ilbIp = "${var.ilbIp}"
+  is_frontend  = true
+  subscription = "${var.subscription}"
+  additional_host_name = "${var.external_host_name}"
+
+  app_settings = {
```
dnenkov: Weird formatting in this file

infrastructure/prod.tfvars
```diff
@@ -0,0 +1,27 @@
+env = "prod"
+deployment_env = "prod"
+
+vault_section = "prod"
+
+packages_environment = "prod"
+packages_version = "2.2.0"
```
dnenkov: version

infrastructure/sandbox.tfvars
```diff
@@ -0,0 +1,33 @@
+env = "sandbox"
+deployment_env = "test"
+
+vault_section = "test"
+
+packages_environment = "test"
+packages_version = "2.2.0"
```
dnenkov: version

infrastructure/variables.tf
```diff
@@ -0,0 +1,288 @@
...
+variable "tenant_id" {
+  description = "(Required) The Azure Active Directory tenant ID that should be used for authenticating requests to the key vault. This is usually sourced from environemnt variables and not normally required to be specified."
+}
+
+variable "client_id" {
+  description = "(Required) The object ID of a user, service principal or security group in the Azure Active Directory tenant for the vault. The object ID must be unique for the list of access policies. This is usually sourced from environment variables and not normally required to be specified."
+}
+
+variable "node_env" {
+  // https://www.dynatrace.com/news/blog/the-drastic-effects-of-omitting-node_env-in-your-express-js-applications/
```
dnenkov: Do we need this line?

package.json
```diff
@@ -4,29 +4,35 @@
...
   },
   "scripts": {
-    "start": "NODE_ENV=production NODE_PATH=. node server",
+    "start": "node server.js",
+    "start-backup": "NODE_ENV=production NODE_PATH=. node server",
     "start-dev": "NODE_ENV=development NODE_PATH=. grunt start-dev",
     "start-debug": "NODE_ENV=development NODE_PATH=. node --inspect=0.0.0.0:9229 server.js",
     "start-debug-ide": "NODE_ENV=development NODE_PATH=. node $NODE_DEBUG_OPTION server.js",
-    "start-prod": "NODE_PATH=. grunt start-prod",
+    "start-prod": "NODE_PATH=. gruntNODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. istanbul cover _mo start-prod",
```
dnenkov: Shouldn't there be a white space in `gruntNODE_ENV=testing`. Also, should this be `testing` or `prod`?

test/end-to-end/codecept.conf.js
```diff
@@ -5,17 +5,18 @@ console.log('waitForAction value set to', waitForAction); // eslint-disable-line
 
 exports.config = {
   'tests': './paths/**/*.js',
-  'output': './output',
+  'output': '../../functional-output',
```
dnenkov: Can we find a way to use absolute urls instead of relative urls that go few levels up

tchow8: Sorry not too sure on how to resolve this. Where the codecept.conf.js is seems to be treated as the config root. Probably is a way though.
Problem is our codecept.conf.js is in test/end-to-end, but its hardcoded into CNP pipeline to check for the output folder (which is exactly named as either smoke-output or functional-output) in the root project folder.

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+selenium-hub:
```
dtaylor7: this is a shame, i have an ignored docker-compose file locally that starts up my local redis...

tchow8: Actually I committed this file for testing, i'll remove now actually thanks for catching. I just made this to prove docker-compose was broken on the jenkins box

.gitignore
```diff
@@ -24,3 +24,7 @@ package-lock.json
 .vscode/launch.json
 .sonar/
 .env
+smoke-output/*
+!smoke-output/.gitkeep
```
dtaylor7: do you need to put rule here if you have .gitkeep?

i personally prefer to have a `.gitignore` file in folder ( i.e. `smoke-output` ) with contents:
```
*
!.gitignore
```

app/middleware/sessions.js
```diff
@@ -68,8 +67,7 @@ const sessions = module.exports = { // eslint-disable-line no-multi-assign
         saveUninitialized: true,
         cookie: {
           secure: cookieSecure,
-          httpOnly: true,
-          domain: hostName
```
dtaylor7: you are not restricting cookies to a domain now?

infrastructure/variables.tf
```diff
@@ -0,0 +1,277 @@
...
+variable "post_code_url" {
+  default = "https://postcodeinfo.service.justice.gov.uk"
+}
+
+variable "hpkp_max_age" {
+  default = "86400"
+}
+
+variable "hpkp_sha256s" {
+  default = "Naw+prhcXSIkbtYJ0t7vAD+Fc92DWL9UZevVfWBvids=,klO23nT2ehFDXCfx3eHTDRESMz3asj1muO+4aIdjiuY=,grX4Ta9HpZx6tSHkmCrvpApTQGo67CYDnvprLg5yRME="
```
dtaylor7: we are happy with keys like this is public repo?

app/middleware/sessions.js
tchow8: I had assumed that express session automatically assigned the current domain to the cookie. if that is not the case, I'll add it back in now using req.get('host')

infrastructure/variables.tf
tchow8: Honestly not sure actually. We didn't store it as a secret (in vault) back in tactical. We would need to move it to vault in order to keep it. Otherwise I can change it to a placeholder value?

dtaylor7: somthing for @dnenkov to approve i think

dnenkov: @tchow8 can we please store that in vault. Also can we change it to a different value as the current value will remain visible in github.com

tchow8:
Opening a new PR to maintain a cleaner history. This new PR should be the final version.

PR #76, opened 2018-04-19
Draftstore restore bug

dtaylor7:
# Description
[DIV-2513 - Save and Resume: Unable to continue -once journey broken](https://tools.hmcts.net/jira/browse/DIV-2513)

A fix to help people that skip questions and navigate directly to jurisdiction (obviously not talking about general public here, just testers ..... ) 

## Type of change

# How Has This Been Tested?

Manually

# Checklist:


dtaylor7:
@dnenkov Thats fine, the blacklisted properties are removed from the session, so in theory we should just be left with the question answers....

dtaylor7:
However, we could still have this problem if a user logs in, navigates to any question but the first one and clicks save and close... we have a scenario then were a user has nothing in there session.... not sure how to fix that one tho

dtaylor7:
@dnenkov @borislavpetrovhmcts i think this is a better solution

PR #80, opened 2018-04-20
Div 2466 form elements must have labels

richard-diamond:
# Description
changed base form elements definition for test fields to include the ability to specify a label class. Changed the foreign marriage certificate page to set labels with hint text for screen readers, with the labels having an invisibility class so they aren't displayed

Fixes # (issue)
DIV-2466, DIV-2504

## Type of change
- [x ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
unit tests, manually progressing through the divorce application story

# Checklist:
- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/views/common/components/formElements.html
+++ b/app/views/common/components/formElements.html
@@ -15,9 +15,18 @@
     </div>
 {% endmacro %}
 
-{% macro textField(name, field, label, hint = '', boldLabel = false) %}
+{% macro textField(name, field, label, hint = '', boldLabel = false, labelClass = '') %}
+
+    {% if labelClass %}
+        {% set actualLabelClass = labelClass %}
+    {% elif boldLabel %}
+        {% set actualLabelClass = "form-label-bold text" %}
+    {% else %}
+        {% set actualLabelClass = "form-label text" %}
+    {% endif %}
+
     <div class="form-group {{ 'form-group-error' if field.error }}">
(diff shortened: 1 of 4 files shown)


app/views/common/components/formElements.html
```diff
@@ -15,9 +15,18 @@
...
+    {% if labelClass %}
+        {% set actualLabelClass = labelClass %}
+    {% elif boldLabel %}
+        {% set actualLabelClass = "form-label-bold text" %}
+    {% else %}
+        {% set actualLabelClass = "form-label text" %}
+    {% endif %}
+
     <div class="form-group {{ 'form-group-error' if field.error }}">
-      <label class="form-label{{ '-bold' if boldLabel }} text" for="{{ name }}">{{ label | safe }}</label>
```
dtaylor7: are you sure we want to remove class here? also removing `-bold` would change display for labels on this page: `app/steps/marriage/certificate-names/template.html`

richard-diamond: I believe we do. The bold isn't lost, the new code placed immediately above that div statement applies the bold class if the boldLabel parameter is true. Moving that logic above the div block allows us to also check for a passed label class, which is something my ticket needs to do to set the label invisible

PR #93, opened 2018-04-25
Fix localhost cookie domain setting

tchow8:
Was setting cookie domains with the whole host (name + port). Cookie domains don't like the port, so am just passing in the hostname part into the session cookie and the idam state cookies.

Just a clarification on the changes I am making:

req.protocol => return 'http' or 'https' etc.
req.hostname => returns the hostname without a port, so 'localhost' or 'divorce.reform.hmcts.net'
req.get('host') => returns the full host, so 'localhost:8080' or 'divorce.reform.hmcts.net' (not sure if it appends the 80/443 actually)
There is no req.port so the way I retrieve port is through req.get('host').split[1]. As any proper hostname will have in hostname:port format, no protocol.
Any hostnames with more than one ':'  is most likely invalid.

### DIFF
--- a/app/middleware/sessions.js
+++ b/app/middleware/sessions.js
@@ -68,7 +68,7 @@ const sessions = module.exports = { // eslint-disable-line no-multi-assign
         cookie: {
           secure: cookieSecure,
           httpOnly: true,
-          domain: req.get('host')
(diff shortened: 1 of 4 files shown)


tchow8:
Opening new PR

PR #105, opened 2018-04-27
Application Submitted Changes

tchow8:
Previous error-application-submitted is not the correct message for when the user tries to start an application when they already have one.
This has been changed to application-submitted.
Additional content has been added to this page.


nsingamsetty:
Tested here https://pr-105-div-pfe-preview.service.core-compute-preview.internal/index and it is working as expected. The Build is ready to Merge.  Please see the attached screenshot for for ticket 2290 for further details

nsingamsetty:
Tested the change and working as expected.

PR #108, opened 2018-04-27
Start app insights before server

tchow8:
# Description

Start application before even loading any other library to make sure app insights captures absolutely everything.


tchow8:
Sandbox, Saat and SProd settings has been merged in. Since preview isn't working I am going to test these new logs on sProd instead.

tchow8:
Opening a new Pr against latest master

PR #132, opened 2018-05-04
 iOS should not recognise case id as phone number

borislavpetrovhmcts:
iOS should not recognise case id as phone number

# Description

Include <meta name="format-detection" content="telephone=no"> on done and submitted step so the case id is not recognised as phone number in iOS.

More documentation about this here : https://developer.apple.com/library/content/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html

Fixes # DIV-1849

## Type of change

# How Has This Been Tested?

**Test Configuration**:

* Hardware:
* O/S and version:
* JDK:

# Checklist:


Rob-Doherty:
I have verified on iPhone and looks correct - iOS no longer tries to auto-dial

PR #146, opened 2018-05-11
Fixed Sandbox, Saat and Sprod as well as starting appInsights before …

tchow8:
…application start


Rob-Doherty:
Possibly we should just remove the `feature-toggle-api`, but it's consistent with the other env .tfvars so can be done as a separate techdebt if necessary

Rob-Doherty:
Merged in https://github.com/hmcts/div-petitioner-frontend/pull/174
Closing.

PR #149, opened 2018-05-14
Fixed session and idam cookie domain issues

tchow8:
Domain can not include the port which is returned with req.get('host').
A valid host will always only contain one colon ( ':' ) so can safely split by this delimiter and extract the zero index to get the actual hostname.


Rob-Doherty:
Merged as part of https://github.com/hmcts/div-petitioner-frontend/pull/174
Closing.

PR #74, opened 2018-04-18
add IDAM userId to logs

dtaylor7:
# Description
[DIV-1931 - Add request / user id to logs from front-end](https://tools.hmcts.net/jira/browse/DIV-1931)
This PR adds the IDAM user id to each log that its possible to be added too.

Note: The base of this branch is `dont-use-idam-jwt` as these changes are required for this update

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Yes visually 

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #151, opened 2018-05-15
DIV-2684 Improve error logging

mattyod:
# Description

Stop coercing errors into template strings and log either the entire
error object or add it as a property on a parent object with message.

Fixes # DIV-2684

## Type of change

# Checklist:

**Before**
screen shot 2018-05-14 at 14 06 00

**After**
screen shot 2018-05-14 at 14 08 44
or
screen shot 2018-05-14 at 14 06 53

**Before**
screen shot 2018-05-14 at 14 40 51

**After**
screen shot 2018-05-14 at 14 36 30


app/middleware/draftPetitionStoreMiddleware.js
```diff
@@ -72,7 +72,7 @@ const removeFromDraftStore = (req, res, next) => {
     .then(next)
     .catch(error => {
       if (error.statusCode !== httpStatus.NOT_FOUND) {
-        logger.error(`Error when attempting to restore session from draft store ${error}`);
```
dtaylor7: would it be more usfull to change these ones as you have done to others? e.g.
```
logger.error({
    message: 'error message...',
    error
});
```

dtaylor7: would mean the upcoming PR https://github.com/hmcts/div-petitioner-frontend/pull/74 would work also

mattyod: In most cases I'm not convinced the message really helps, often it could be a number of things that have gone wrong. i.e code error vs http error.

In the examples screen-shotted you now get a file and line number from the stack trace, as well as any message already attached to the error, which I think is more useful.

In certain cases such as health checks, custom messages perhaps make more sense.

I don't think every error log has to be the same, whatever best enables debugging.

dtaylor7: Fair enough, however in the PR i mentioned we are attaching the IDAM id to the message. I suppose that could be edited to create a message attribute instead of modify current one tho...

PR #166, opened 2018-05-21
Moved court allocation to submission and refresh court data before pa…

tchow8:
[DIV-2707](https://tools.hmcts.net/jira/browse/DIV-2707)

This PR moves court allocation from the marriage certificate upload page to the submission page.
Previously, the user would see the court details when they select 'Can not upload'. This has been changed so the user will see the court details to send their documentation to on the Done or Done-And-Submitted pages only. as such, court allocation is no longer required to occur at this time.

What is believed to be the issue for the above JIRA ticket is that some users who Save and Quit their application after court allocation before the initial Site ID fix came in saved the incorrect Site ID to their session. As such when they eventually came back to pay, their payment was associated with the wrong Site ID.

This PR assigns a new court and the court details at submission only. The reasoning is that submission should only ever occur once per user, meaning that if they submit a case, it should be allocated a court (and that even if the session already had an allocated court, that would simply be old data that can be replaced).
This PR also populates the court details (which contain the current siteID) before payment, where Site ID is used to populate the payment details.

The population of court details is required at Submission time to be visible on the Done page (with help with fees active).
Then the re-population occurs before actual payment since the user can Save and Quit after submission but before payment. The court ID is controlled by us and we can stop that from changing, but the court details such as address and site ID could easily be updated and this allows the correct SiteID to be used at the point of payment.


Rob-Doherty:
Merging on @nsingamsetty 's behalf

nsingamsetty:
This is tested in Preview and working as expected.

https://pr-166-div-pfe-preview.service.core-compute-preview.internal/

Here are the following cases for reference which are tested against this ticket.

1527 - 1809 - 3428 - 7907-AA03

1527 - 1793 - 3390 - 3676-AA01

1527 - 1723 - 1265 - 1804-AA04

1527 - 1721 - 1701 - 9207-AA02

Build successful - https://build.platform.hmcts.net/view/all/job/HMCTS/job/div-petitioner-frontend/view/change-requests/job/PR-166/

PR #167, opened 2018-05-22
DIV-2485 Prevent case reference number from vanishing when user clicks 'back' button from Policy Privacy page

wjtlopez:
Returning initSession as part of get middleware in DestroySessionStep so that the user is redirected to the index page when there is no session


nsingamsetty:
Tested here. https://pr-167-div-pfe-preview.service.core-compute-preview.internal/index. PR build successful. As per the Jira Ticket this is working as expected hence ready to merge. changing the status of the DIV-2485 to done once merge successful.

PR #174, opened 2018-05-25
Multiple TechDebt PR's in one

dtaylor7:
# Description

This PR includes the following PR's:
 - [Fixed session and idam cookie domain issues](https://github.com/hmcts/div-petitioner-frontend/pull/149)
 - [Use local codeceptjs instead of global one provided by the docker image](https://github.com/hmcts/div-petitioner-frontend/pull/153)
 - [clean up of vault creation code](https://github.com/hmcts/div-petitioner-frontend/pull/100)
 - [DIV-2178: changed to use the cloned context to store the next step ur…](https://github.com/hmcts/div-petitioner-frontend/pull/127)
 - [Fixed Sandbox, Saat and Sprod as well as starting appInsights before](https://github.com/hmcts/div-petitioner-frontend/pull/146)

Please see above links for details on these changes


server.js
```diff
@@ -1,3 +1,12 @@
+const appInsights = require('applicationinsights');
+const CONF = require('config');
+
+if (CONF.environment === 'production') {
```
Rob-Doherty: This won't work as `'production'` is not configured anywhere.
Spoke with @tchow8, who's suggested just checking for the presence of the key instead (so that it always works in CNP)

tchow8: To clarify:
if (CONF.applicationInsights.instrumentationKey) {
}
As instrumentationKey will be undefined unless set by terraform.

Rob-Doherty:
Verified app insights working for branch, regression proves refactors still working, all e2e tests pass. Ready to merge.

PR #205, opened 2018-06-13
DIV-2794 Do not run app.listen() for unit tests

mattyod:
# Description

Frequently seeing the 'test' step of pipeline build fail due to:

`Uncaught Error: listen EADDRINUSE :::3001`

This is likely due to the asynchronous nature ofthe Node.js `http.close()` method but it's not necessary for our Supertest.js tests to initialise listening for connections, so simplest and safest not to initialise `app.listen()` when testing that way.

Fixes # DIV-2794

# Checklist:

**n.b.** tests may still fail in other difficult to reproduce ways but this is part of a wider epic.

### DIFF
--- a/app.js
+++ b/app.js
@@ -42,7 +42,7 @@ const PORT = process.env.PORT || process.env.HTTP_PORT || CONF.http.port;
 
 const logger = logging.logger(__filename);
 
-exports.init = () => {
+exports.init = listenForConnections => {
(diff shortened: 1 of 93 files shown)


Rob-Doherty:
Have been able to do full end-to-end regression journey including CCD and all looks fine. ready to merge. 👍

PR #217, opened 2018-06-20
Remove feature toggle healthcheck from demo

tchow8:
This removes the feature toggle healthcheck from demo and brings in line with latest master


tchow8:
Closing as not required

PR #221, opened 2018-06-21
Comment out url for feature-toggle-api and use configs

tchow8:
Temporary quick fix so that we don't get issues when tactical env goes does again.

Removes feature-toggle-api healthcheck (since we ave commented out using that url) and also changes redis url from tactical to cnp.


Rob-Doherty:
All e2e tests are green (ran locally against the preview instance), manual exploratory has been successful, all toggles appear to be working as intended

PR #233, opened 2018-06-28
Changing title based on user feedback that it was confusing.

NeilMcW:
Minor change to fix a potentially confusing question.


Rob-Doherty:
verified that this won't impact the frontend auto tests

PR #249, opened 2018-07-12
Fix bug where confirmPrayer was not being added to the session

tchow8:
[DIV-2979](https://tools.hmcts.net/jira/browse/DIV-2979)

D8StatementOfTruth which is mapped from DivorceSession.confirmPrayer was not being set in CCD. The reason was a change which no longer did submission after the CYA but as part of it, meaning the confirmPrayer variable was never added to the session before submission to CCD, but only after it, meaning it wasn't saved.
This would mean D8StatementOfTruth is null in CCD for these scenarios.
The second scenario where D8StatementOfTruth is set to 'No' would occur if the user did a Save and Close on CYA and then resumed and submitted.

The change basically adds in adding the current ctx to the session before the submission occurs.

### DIFF
--- a/app/steps/check-your-answers/index.js
+++ b/app/steps/check-your-answers/index.js
@@ -85,6 +85,24 @@ module.exports = class CheckYourAnswers extends ValidationStep {
     ];
   }
 
+  parseRequest(req) {
+    const ctx = super.parseRequest(req);
+
+    // if confirmPrayer has no value, set it to false
+    ctx.confirmPrayer = ctx.confirmPrayer || false;
+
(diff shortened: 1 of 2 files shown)


Rob-Doherty:
We've not been able to re-build the PR in preview due to CNP issues, however we have been able to sign-off the cases generated above with the Business and all seems fine.

In this case we've taken a 'risk-based approach' and merged the change that has been exploratory tested and signed off into master, where the full e2e test suite will be run in AAT instead. The risk is that any failures could potentially block route to production / dirty master branch, but this can be reverted if needed. Given the severity of the issue this seems a fair risk.

PR #251, opened 2018-07-13
Reduce e2e setup for Payments e2e tests

Rob-Doherty:
# Description

This change introduces a new way to set `basicDivorceSessionData.js` default session for any e2e test (in this case we're starting with `payment.js`) so that setup repetition is avoided. This makes tests faster and more atomic.

Also, the `basicDivorce.js` test has been updated to validate the session data generated during the full end-to-end journey against this static data to ensure any divergence between it and the app are caught. **_THIS IS EXTREMELY IMPORTANT_** as it ensures any tests that rely on that static data can still be trusted.

So now in other tests we can use 
```javascript
  I.startApplicationWithAnExistingSession();
```
instead of 
```javascript
  I.amOnLoadedPage('/index');
  I.startApplication();
  I.haveBrokenMarriage();
  I.haveRespondentAddress();
  I.haveMarriageCert();

  I.selectHelpWithFees();
  I.enterHelpWithFees();
  I.selectDivorceType();
  I.enterMarriageDate();
  .
  .
```
and configure the important parts relevant for that test, for example remove the HWF choice
```javascript
  I.startApplicationWithAnExistingSession();
  I.amOnLoadedPage('/pay/help/need-help');
  I.selectHelpWithFees(false);
  I.amOnLoadedPage('/check-your-answers');
  I.checkMyAnswers();
```

Please Note: The `/session` API has been exposed in test envs only so that the tests can GET and SET the user's session.

Fixes # DIV-2737

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Test only change

# How Has This Been Tested?

This has been tested locally and will be confirmed on PR instance with full auto pack before merging.

**Test Configuration**:

* Hardware:
* O/S and version:
* JDK:

# Checklist:


test/end-to-end/data/basicDivorceSessionData.js
```diff
@@ -0,0 +1,218 @@
+module.exports = {
```
dtaylor7: is this not same file in /test/fixtures/mockSession.json?

test/end-to-end/helpers/SessionHelper.js
```diff
@@ -0,0 +1,113 @@
...
+      cookieHeaders.Cookie = `${cookieHeaders.Cookie}; ${authTokenCookie.name}=${authTokenCookie.value}`;
+    }
+
+    return await new Promise((resolve) => {
+      unirest.get(`${helper.config.url}/session`)
+        .headers(cookieHeaders)
+        .strictSSL(false)
+        .proxy(proxyUrl)
+        .end((response) => {
+          // console.log('### SESSION GET:', response.body);
```
dtaylor7: remove

test/end-to-end/helpers/SessionHelper.js
```diff
@@ -0,0 +1,113 @@
...
+    const connectSidCookie = await helper.grabCookie('connect.sid');
+    const authTokenCookie = await helper.grabCookie('__auth-token');
+    const session = await this.getTheSession(connectSidCookie, authTokenCookie);
+
+    let basicSession = this.updateExpectedSessionWithActualSession(basicDivorceSessionData, session);
+
+    await this.setTheSession(connectSidCookie, authTokenCookie, basicSession);
+  }
+
+  updateExpectedSessionWithActualSession(expectedSession, actualSession) {
```
dtaylor7: not best just to do `Object.assign(expectedSession, actualSession);`?

test/end-to-end/data/basicDivorceSessionData.js
Rob-Doherty: mockSession.json looks to contain all possible key / value pairs that the service can generate as session data, including combinations that aren't possible from the frontend (e.g. all 5 reasons for divorce). So I don't think it's going to be worthwhile trying to wedge the 2 different data sets into 1.

test/end-to-end/helpers/SessionHelper.js
Rob-Doherty: will do

test/end-to-end/helpers/SessionHelper.js
Rob-Doherty: I'm trying to re-use this method in 2 places: one to setup the static (expectedSession) data to be POSTed into the user's session (which would suit your suggestion), and one to setup the static data ready for assertion against the user's actual session. So in the 2nd one I want to make sure only specific fields are updated, not all of them otherwise the comparison check would be pointless.
If you'd rather me use `Object.assign()` for the 1st requirement and leave the 2nd requirement as is i'm happy to do that, though not sure if it gives us any benefit?

Rob-Doherty:
@tchow8 does the `logger` actually output anything in Codecept logs though? is that something we can configure?

tchow8:
It should do, as long as the log level is set correctly then it'll output to the same place console.log usually would. But the control is for example you use log.info for saying Idam User Created. Then you use log.debug for the session data.
If before the test you set LOG_LEVEL=info, you'll only get the Idam User Created message and that way you can skip a lot of the session data outputted (which isn't sensitive in a test case but will pollute the logs, but its not important enough to request a change imo).
But if you want to check your session, just set LOG_LEVEL=Debug and it should just output without needing to change any code.
Then if Codecept logs are capturing anything that gets outputted by console.log, normally it'll capture anything that gets logged by logger too.

Rob-Doherty:
ah cool, thanks for the clarification @tchow8, I've updated as per your suggestion

Rob-Doherty:
Please note: I've also ran the full e2e test set against this PR and it all passes.

dtaylor7:
Could you re-enable e2e test on build so we can make sure they work on there? ( can disabled them before merge )

PR #257, opened 2018-07-20
testing-str-of-user-resume-bug

johnoneill46:
Solely for testing purpose

Bug could be due to multiple cases in CCD & so could be fixed when we implement multiple cases in CCD story


tchow8:
Test Pr that is too old now. Closing

PR #258, opened 2018-07-20
Potential fix for bug dont overwrite this one rob

johnoneill46:


dtaylor7:
this could be the problem... https://github.com/hmcts/div-petitioner-frontend/blob/master/app/steps/check-your-answers/index.js#L275

PR #261, opened 2018-07-26
fix for the fee amount on redis

borislavpetrovhmcts:
fix for the fee amount on redis

### DIFF
--- a/app/middleware/updateApplicationFeeMiddleware.js
+++ b/app/middleware/updateApplicationFeeMiddleware.js
@@ -39,7 +39,7 @@ const getFeeFromFeesAndPayments = () => {
 };
 
 const updateApplicationFeeMiddleware = (req, res, next) => {
-  redisClient.get('commonProps.applicationFee')
+  redisClient.get('commonProps.applicationFee.amount')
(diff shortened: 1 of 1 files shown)


app/middleware/updateApplicationFeeMiddleware.js
```diff
@@ -39,7 +39,7 @@ const getFeeFromFeesAndPayments = () => {
 };
 
 const updateApplicationFeeMiddleware = (req, res, next) => {
-  redisClient.get('commonProps.applicationFee')
+  redisClient.get('commonProps.applicationFee.amount')
```
Rob-Doherty: could a unit test be added, to prove the fix?

tchow8:
Just having some discussion on improving the unit tests for this.
Maybe changing stuff like:
 expect(ioRedisClient.get.calledOnce).to.eql(true);
to expect(ioRedisClient.get.calledWith).to.eql('commonProps.applicationFee.amount')
and some other stuff that you can think of.

borislavpetrovhmcts:
May be create tech debt ticket about this ?

PR #267, opened 2018-07-27
[PREVIEW] Fixed the links to go back to reason for divorce on desertion exit pages

tchow8:
[DIV-2750](https://tools.hmcts.net/jira/browse/DIV-2750)

This PR fixes the html links on desertion exit pages redirecting to a now non-existent page.


tchow8:
No test, as this is essentially a content change. Although up for discussion at what level we test at. Otherwise instead of using an anchor link we should control the logic of the back via the node server.

nsingamsetty:
This is tested and working as Expected in the below PR.
https://pr-267-div-pfe-preview.service.core-compute-preview.internal/index

Full E2E tests ran successfully with out any issues.Please refer DIV-2750 for evidences. Merging this PR Now.

PR #282, opened 2018-08-10
[PREVIEW] Use waitInUrl instead of seeCurrentUrlEquals

tchow8:
Our existing seeCurrentUrlEquals which is a codecept method is being intercepted in the JSWait Helper class and waits for the DOM Element #content before proceeding.
Gov.UK Payment page recently changed the id of that element to #maincontent instead meaning our E2E tests no longer  find what we are expecting and the test times out and fails.
This PR changes the seeCurrentUrlEquals to a WaitInUrl (also codecept function) that does a similar thing to seeCurrentUrlEquals but with improved checking for page load logic, but more importantly does not check for a DOM element that isn't in our control.


nsingamsetty:
This is Tested and executed all the e2e functional tests against PR and it is working as expected. Merging this PR

PR #287, opened 2018-08-22
Cleaned up local vault naming

tchow8:
master build is failing for some vault unsupported protocol issue. This PR is to try to resolve it.


tchow8:
Added back in provider block when I assumed was removable. Seems like it doesn't have an impact on Preview envs, but broke on AAT.

nsingamsetty:
Merging this PR to fix vault issues facing with master builds

PR #292, opened 2018-08-23
[PREVIEW] Redirect to application submitted on back to submit page

tchow8:
Users can go back after submitting and re submit and create a duplicate case.

https://tools.hmcts.net/jira/browse/DIV-2807


dtaylor7:
https://github.com/hmcts/div-petitioner-frontend/blob/master/app/core/steps/ValidationStep.js#L31
https://github.com/hmcts/div-petitioner-frontend/blob/master/app/middleware/submissionMiddleware.js#L3

PR #294, opened 2018-08-24
Parse the boolean for the flag so even strings resolve correctly

tchow8:
When infrastructure variables are taken in by config, they are converted to String. Locally this is the case as well when trying to pass in an env variable such as FEATURE_REDIRECT_TO_APPLICATION_SUBMITTED=false, the code treats that as boolean true strangely. Passing is as false directly to the default.yml does not have this issue.

This code change parses the passed in parameter and if string, checks its equivalent to a string 'true' or if not, to the Boolean true.

### DIFF
--- a/app/core/utils/parseBool.js
+++ b/app/core/utils/parseBool.js
@@ -1,3 +1,3 @@
 module.exports = (bool = '') => {
-  return typeof bool === 'string' ? bool.toLowerCase() === 'true' : bool === true;
+  return String(bool) === 'true';
 };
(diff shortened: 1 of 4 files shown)


app/core/utils/parseBool.js
```diff
@@ -0,0 +1,3 @@
+module.exports = (bool = '') => {
+  return typeof bool === 'string' ? bool.toLowerCase() === 'true' : bool === true;
```
dtaylor7: could be
```
return String(a) == "true";
```

tchow8:
Cheers guys, I've updated the code if you an re approve :) I've tested all again:
@dtaylor7 @borislavpetrovhmcts @qzhou-hmcts

PR #298, opened 2018-08-29
[PREVIEW] Use defined asp_name and asp_rg

tchow8:
Required for the upcoming shared resource group change.


tchow8:
Do this as a proper story.

PR #300, opened 2018-08-30
Fix for prod not using feature flag

tchow8:
I believe the problem being. We take these values as true: true, "true", 1
We were passing to Azure true. It was converting it 1. But then teh configs convert it to "1"


tchow8:
Required as priority to fix build, so I am merging. Tech related, no test required.

PR #306, opened 2018-09-07
change wait

vianvi:


vianvi:
passed from local connection to saucelab
screen shot 2018-09-07 at 11 56 00

PR #314, opened 2018-09-20
Setup shared app service plan settings

tchow8:


tchow8:
Reopening with PREVIEW

PR #330, opened 2018-10-15
[PREVIEW] court allocation

mthava:
# Description

Court allocation based on the reason for divorce

Fixes # (issue)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

**Test Configuration**:

* Hardware:
* O/S and version:
* JDK:

# Checklist:


app/services/courtsAllocation.js
```diff
@@ -1,14 +1,86 @@
...
-    }, {});
+const caseDistribution = CONF.commonProps.divorceFactsRatio;
+const courts = CONF.commonProps.court;
 
-  return weighted.select(options);
+const weightPerFactPerCourt = {};
+const allocationPerFactLeft = {};
+const remainingWeightForCourt = {};
+
+const calculatePreAllocations = () => {
```
dtaylor7: this is hard to read. Can you refactor, suggest splitting into small more bespoke functions with descriptive function names 

name your conditional statements e.g.
```
const divorceFactUndefined = courts[courtName].divorceFactsRatio && typeof courts[courtName].divorceFactsRatio[fact] !== 'undefined';
if (divorceFactUndefined) { 
   ...
```

also you shouldn't need to check typeof, `courts[courtName].divorceFactsRatio && courts[courtName].divorceFactsRatio[fact]` is the same

mthava: I have refactored the method and moved the code into really small methods. Hopefully, this will make it easier to understand the logic.

the problem with if(courts[courtName].divorceFactsRatio) as if(0) return false. In my case if(0) is a valid case. 

Can you please review this again

app/services/courtsAllocation.js
```diff
@@ -1,14 +1,113 @@
...
+  if (remainingWeightForCourt[courtName] < 0) {
+    throw new Error(`Total weightage exceeded for court ${courtName}`);
+  }
+};
+
+const updateAllocationRemainingForFact = (fact, courtName) => {
+  allocationPerFactLeft[fact] -= courts[courtName].divorceFactsRatio[fact];
+
+  if (allocationPerFactLeft[fact] < 0) {
+    throw new Error(`Total weightage exceeded for fact ${fact}`);
```
dtaylor7: how are you handling these errors?

app/services/courtsAllocation.test.js
```diff
@@ -1,44 +1,184 @@
...
+    CONF.commonProps = {
+      divorceFactsRatio: caseDistribution,
+      court: localCourts
     };
-    const expected = {
-      foo: 1,
-      bar: 2
+
+    try {
+      require(modulePath);
```
dtaylor7: i doubt this test works, when you require something its put into cache so the next require just reads from the cache. The only way to test this correctly is to use requireNoCache from one per page test suite see https://github.com/hmcts/div-decree-nisi-frontend/blob/2cea35d93ab47d4e20539b43882f66241c181231/test/unit/services/idam.test.js#L5

config/default.yaml
```diff
@@ -70,6 +70,7 @@ commonProps:
     version: 4
     amount: 550.00
   financialOrderApplicationFee: 255
+  divorceFactsRatio: { "unreasonable-behaviour": 0.30, "separation-2-years": 0.37, "separation-5-years": 0.21, "adultery": 0.11, "desertion": 0.01}
```
dtaylor7: this should be:
```
divorceFactsRatio:
  unreasonable-behaviour: 0.30,
  separation-2-years: 0.37
  separation-5-years: 0.21
...
```

dtaylor7: to keep coding style

app/services/courtsAllocation.test.js
mthava: check the aftereach. I am clearing the cache

app/services/courtsAllocation.js
mthava: this will break application deployment. this is what expected

dtaylor7: that cant be the right approach?

dtaylor7: there should be a comment to explain why

dtaylor7: and what is expected

config/default.yaml
mthava: I really dont want split them into separate variables in terraform it is one block data.

dtaylor7: i understand but thats how its done in in config, when your long gone, someone will come to this and be confused why its different, perhaps question it, but if looks the same there is no issue.

This is one of the reasons we have coding style and try to keep everything the same

mthava: I am not sure. I feel splitting these fields into separate block was wrong in the first place. All the court config should have been one block of data as they are one configuration. If we are to fix it then we need to fix the old config. By keeping them in one block the whole thing can be data-driven. And when a new court gets introduced we shouldn't really touch the code.

@tchow8 @mrganeshraja what is ur thought on this

dtaylor7: this is to do with code style and standards, keep the code style the same through app.

app/services/courtsAllocation.js
mthava: added comments

dtaylor7:
👍

PR #336, opened 2018-10-26
WIP Div app router

tchow8:
Testing routing


tchow8:
After discussions, a better way was proposed. Closing.

PR #340, opened 2018-11-02
Redirect to DN if no case or not AwaitingPayment

tchow8:
This is the Petitioner Redirect functionality on any Question page, running after idam.protect checks its a valid user. User session should be populated at this time so it can check for a case.
If no case or case state is in AwaitingPayment it will just call next(), otherwise will redirect to the passed in DECREE NISI FRONTEND URL which is set in an environment variable.

This is the first of three parts, requiring a PR on DN side for the opposite case and then ensuring auth cookies share the same domain.


app/middleware/redirectMiddleware.js
```diff
@@ -0,0 +1,13 @@
+const CONF = require('config');
+
+const redirectOnCondition = (req, res, next) => {
+  // If state, there must be a CCD case
+  const hasCaseState = req.session && req.session.state;
+  if (hasCaseState && !CONF.ccd.d8States.includes(req.session.state)) {
+    return res.redirect(CONF.apps.dn.url);
```
mthava: this will fail when you go live. On live you will need to use public DNS additional host name

mthava:
And missing AKS changes as well

config/default.yaml
```diff
@@ -192,3 +192,11 @@ saucelabs:
...
 public:
   protocol: 'https'
+
+apps:
+  dn:
+    url: 'https://localhost:3000'
+
+ccd:
+  d8States:
+    - 'AwaitingPayment'
```
mthava: you may need to add submitted state here

app/middleware/redirectMiddleware.js
mthava: I see you have set the dn url as public dns in prod.tf.

config/default.yaml
tchow8: Will we still go to PFE with Submitted? I thought since they would be unable to do anything on Frontend with Submitted state, they should see the progress-bar page instead?

PR #350, opened 2018-11-22
Adding new court to list of courts

jvictorcf:
With this alteration we should be ready to start assigning cases with the reason of "unreasonable behaviour" to the CTSC.
Also, I've added the template changes to be able to render the new court format correctly.

# Description

This is the first step to adding the CTSC court to the list of available courts.

## Type of change

# How Has This Been Tested?

App's behaviour has not changed as part of this PR, therefore existing tests passing should be enough assurance that existing behaviour keeps working.

# Checklist:


jvictorcf:
Hello Tony,

I’ll be doing the “c/o” as part of the rendering rather than having it in the data storage. Yes, the COS one has been done already.
Is that ok for you?

tchow8:
Yep fine, approved

config/custom-environment-variables.yaml
```diff
@@ -91,7 +91,19 @@ commonProps:
...
+      divorceCentre: COURT_SERVICE_CENTRE_NAME
+      courtCity: COURT_SERVICE_CENTRE_CITY
+      poBox: COURT_SERVICE_CENTRE_STREET
+      postCode: COURT_SERVICE_CENTRE_POSTCODE
+      openingHours: COURT_SERVICE_CENTRE_OPENINGHOURS
+      email: COURT_SERVICE_CENTRE_EMAIL
+      phoneNumber: COURT_SERVICE_CENTRE_PHONENUMBER
+      siteId: COURT_SERVICE_CENTRE_SITEID
+      weight: COURT_SERVICE_CENTRE_WEIGHT
+      divorceFactsRatio: COURT_SERVICE_CENTRE_DIVORCE_FACT_RATIO
```
borislavpetrovhmcts: Do you need this ? I think 'Service centre' should never be allocated (Check this with BA). Can you check how the court allocation works and if needed, make sure it is never allocated - make it always 0 or exclude it somehow.

config/custom-environment-variables.yaml
```diff
@@ -91,7 +91,19 @@ commonProps:
...
+      serviceCentreName: SERVICE_CENTRE_NAME
+      divorceCentre: COURT_SERVICE_CENTRE_NAME
+      courtCity: COURT_SERVICE_CENTRE_CITY
+      poBox: COURT_SERVICE_CENTRE_STREET
+      postCode: COURT_SERVICE_CENTRE_POSTCODE
+      openingHours: COURT_SERVICE_CENTRE_OPENINGHOURS
+      email: COURT_SERVICE_CENTRE_EMAIL
+      phoneNumber: COURT_SERVICE_CENTRE_PHONENUMBER
+      siteId: COURT_SERVICE_CENTRE_SITEID
+      weight: COURT_SERVICE_CENTRE_WEIGHT
```
borislavpetrovhmcts: Do you need this ? I think 'Service centre' should never be allocated (Check this with BA). Can you check how the court allocation works and if needed, make sure it is never allocated - make it always 0 or exclude it somehow.

borislavpetrovhmcts: Do you need this ? I think 'Service centre' should never be allocated (Check this with BA). Can you check how the court allocation works and if needed, make sure it is never allocated - make it always 0 or exclude it somehow.

jvictorcf: As per our conversation, this is the business understanding.

jvictorcf:
I've made further changes to this PR. Now unit tests were added following the same logic of the existing tests and the CTSC address is rendered correctly.
Could you please have another look into it and see if you're still happy with it @tchow8 and @borislavpetrovhmcts ?

app/steps/done-and-submitted/index.test.js
```diff
@@ -397,6 +419,59 @@ describe(modulePath, () => {
...
+      withSession(done, agent, session);
+    });
+
+    serviceCentreDetails.forEach(courtDetail => {
+      it(`contains serviceCentre contact detail: ${courtDetail}`, done => {
+        testExistence(done, agent, underTest, courtDetail);
+      });
+    });
+
+    it(`contains serviceCentre contact detail: ${CONF.commonProps.court.serviceCentre.openingHours}`, done => {
```
hosainnet: We could `openingHours` in `serviceCentreDetails` and remove this test

infrastructure/prod.tfvars
```diff
@@ -25,3 +25,7 @@ decree_nisi_frontend_url = "https://dn.apply-divorce.service.gov.uk"
 capacity = "2"
 
 feature_redirect_on_state = "false"
+
+court_service_centre_divorce_facts_ratio = ""
```
hosainnet: Is this to effectively disable it in production? Is it safe to be `""`?

infrastructure/preview.tfvars
```diff
@@ -24,4 +24,6 @@ feature_redirect_to_application_submitted = "false"
 
 dev_support_notification_email = "divorce-team-a@hmcts.net"
 
-feature_redirect_on_state = "false"
\ No newline at end of file
+feature_redirect_on_state = "false"
+
+court_service_centre_divorce_facts_ratio = ""
```
jvictorcf: Should this be here, @mthava?

mthava: this is just for testing. Once you have tested this on PREVIEW envoirnment remove it

app/steps/done-and-submitted/index.test.js
jvictorcf: Good catch!

jvictorcf: Done!

infrastructure/preview.tfvars
jvictorcf: Done!

infrastructure/main.tf
```diff
@@ -215,17 +215,30 @@ module "frontend" {
...
+    COURT_SERVICE_CENTRE_PHONENUMBER        = "${var.court_service_centre_phonenumber}"
+    COURT_SERVICE_CENTRE_SITEID             = "${var.court_service_centre_siteid}"
+    COURT_SERVICE_CENTRE_WEIGHT             = "${var.court_service_centre_weight}"
+
     COURT_EASTMIDLANDS_DIVORCE_FACT_RATIO   = "${var.court_eastmidlands_divorce_facts_ratio}"
     COURT_WESTMIDLANDS_DIVORCE_FACT_RATIO   = "${var.court_westmidlands_divorce_facts_ratio}"
     COURT_SOUTHWEST_DIVORCE_FACT_RATIO      = "${var.court_southwest_divorce_facts_ratio}"
-    COURT_SOUTHWEST_DIVORCE_FACT_RATIO      = "${var.court_northwest_divorce_facts_ratio}"
+    COURT_NORTHWEST_DIVORCE_FACT_RATIO      = "${var.court_northwest_divorce_facts_ratio}"
+    COURT_SERVICE_CENTRE_DIVORCE_FACT_RATIO = "${replace(jsonencode(var.court_service_centre_divorce_facts_ratio), "/\"([0-9]*\\.?[0-9]*)\"/", "$1")}"
```
dtaylor7: why `replace(jsonencode...`?

jvictorcf: This bit was done by Mathan. I believe the replace change double-quotes for single-quotes and the jsonencode transforms a map into a json.

infrastructure/prod.tfvars
```diff
@@ -25,3 +25,11 @@ decree_nisi_frontend_url = "https://dn.apply-divorce.service.gov.uk"
 capacity = "2"
 
 feature_redirect_on_state = "false"
+
+court_service_centre_divorce_facts_ratio = {
```
alexwakeman: Curious about why we do this. Possibly it's because we don't do weightings in production yet.

app/services/courtsAllocation.js
```diff
@@ -9,6 +9,10 @@ const weightPerFactPerCourt = {};
 const allocationPerFactLeft = {};
 const remainingWeightForCourt = {};
 
+function _getDivorceFactRatioForCourt(courtName, fact) {
+  return parseFloat(courts[courtName].divorceFactsRatio[fact]);
```
alexwakeman: Could consider returning a default value of say `0` if the fact is invalid for examaple - just a suggestion.

jvictorcf: I understand where you're coming from, but I'm not sure 0 is the default value or if we'd rather break the whole thing if configuration is wrong.
I think that this whole court allocation thing needs to be refactored/moved into our REST API and when we do this, will be the ideal time to discuss the behaviour the business wants.

I'm going to create a story for this. Thanks for the comment.

infrastructure/prod.tfvars
jvictorcf: In prod, CTSC shouldn't deal with any cases.

jvictorcf: In prod, CTSC shouldn't deal with any cases.

PR #354, opened 2018-11-30
Fix healthchecks

tchow8:
Passing in options as an object in the options object doesn't work, it must be part of the healthcheck options object itself. This should mean the timeouts take effect from config now, so 2s instead of 5s.

This should fix our false positive healthcheck timeouts as cos can take up to 3s to return, lower than the 2s max.
Our config already had 5s as the timeout, but it wasn't being used correctly as it was passed into the node healthcheck incorrectly. This PR fixes that. Currently testing in sprod.

### DIFF
--- a/app/services/healthcheck.js
+++ b/app/services/healthcheck.js
@@ -19,9 +19,17 @@ client.on('error', error => {
   logger.error(error);
 });
 
-const options = {
-  timeout: config.health.timeout,
-  deadline: config.health.deadline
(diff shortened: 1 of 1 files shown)


tchow8:
Direct unit tests don't exist for this yet. Possibly may raise a techdebt to resolve this if coverage falls too much.

nsingamsetty:
This is verified and merging this PR.

PR #285, opened 2018-08-16
Fix logout for sidam

tchow8:
*** NOTE: REQUIRED PRIOR TO SIDAM RELEASE ***

SIDAM requires authentication for logout.

Frontend was missing the client_id and required a change to be made in idam-express-middleware. This PR adds in the client_id to be passed from the infrastructure (was using the default.yml from config) and uses the new idam-express-middleware version.

PR #361, opened 2018-12-06
Added missing IDAM Client-ID required for SIDAM functionality

tchow8:
*** NOTE: REQUIRED PRIOR TO SIDAM RELEASE ***

SIDAM requires authentication for logout.

Frontend was missing the client_id and required a change to be made in idam-express-middleware. This PR adds in the client_id to be passed from the infrastructure (was using the default.yml from config) and uses the new idam-express-middleware version.

Updated version of PR #285


infrastructure/main.tf
```diff
@@ -115,6 +115,7 @@ module "frontend" {
     IDAM_LOGIN_URL                     = "${var.idam_authentication_web_url}${var.idam_authentication_login_endpoint}"
     IDAM_AUTHENTICATION_HEALHCHECK_URL = "${var.idam_authentication_web_url}${var.health_endpoint}"
     IDAM_SECRET                        = "${data.azurerm_key_vault_secret.idam_secret.value}"
+    IDAM_CLIENT_ID                     = "${var.idam_client_id}"
```
hosainnet: Should we add this to the K8S config too?

tchow8: Good catch

tchow8: done

PR #119, opened 2018-04-30
Cya textwrap

dtaylor7:
# Description
[DIV-1448 - Text not wrapping correctly on CYA](https://tools.hmcts.net/jira/browse/DIV-1448)
Remove css rules that are breaking the words on CYA page

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Manually

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #369, opened 2018-12-17
DIV-3815 Apply style change for Firefox -> CYA page

alexwakeman:
# Description

Word-break being applied to td element messes up text flow only in Firefox. This removed.

Fixes # DIV-3815

## Type of change

# How Has This Been Tested?

Tested in FF and Chrome as working.

# Checklist:

### DIFF
--- a/app/assets/sass/_check-your-answers.scss
+++ b/app/assets/sass/_check-your-answers.scss
@@ -1,5 +1,10 @@
 .check-your-answers {
 
+  table.listing {
+    table-layout: fixed;
+    width: 100%;
+  }
+
   th {
     vertical-align: top;
     width: 25%;
@@ -8,7 +13,6 @@
 
   td {
     vertical-align: top;
-
     overflow-wrap: break-word;
     word-wrap: break-word;
   }
--- a/app/assets/sass/_check-your-answers.scss
+++ b/app/assets/sass/_check-your-answers.scss
@@ -3,28 +3,28 @@
   table.listing {
     table-layout: fixed;
     width: 100%;
-  }
 
-  th {
-    vertical-align: top;
-    width: 25%;
-    font-weight: normal;
-  }
+    th {
+      vertical-align: top;
+      width: 25%;
+      font-weight: normal;
+    }
 
-  td {
-    vertical-align: top;
-    overflow-wrap: break-word;
-    word-wrap: break-word;
-  }
+    td {
+      vertical-align: top;
+      overflow-wrap: break-word;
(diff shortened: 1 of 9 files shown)


app/assets/sass/_check-your-answers.scss
```diff
@@ -8,11 +8,6 @@
 
   td {
     vertical-align: top;
-    -ms-word-break: break-all;
-    word-break: break-all;
-
-    /* Non standard for WebKit */
-    word-break: break-word;
```
dtaylor7: this seems dangerous. I remember putting this in to fix a bug months ago

alexwakeman: Ok fair enough - I tested it on Chrome and it worked fine. Any idea which bug it fixed? It is definitely breaking the layout on FF.

dtaylor7: just trying to find out, found the commit https://github.com/hmcts/div-petitioner-frontend/commit/ef9f3e76c5613df47b471e057e1bf851314feb1f

dtaylor7: just not sure how to find the PR

dtaylor7: got it: https://github.com/hmcts/div-petitioner-frontend/pull/119

dtaylor7: As i remember it ( should have left a comment ). With some IE browsers break-word was not working (https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#Browser_compatibility). So break-all is used as a fallback, otherwise the layout of the table would break ( looking very bad ).

So ideally we would find a way to break-word on firefox without breaking functionality on certain IE devices

alexwakeman: @dtaylor7 tested on IE8, IE9, Edge, FireFox, Chrome and iOS Safari and is working as expected.

```
overflow-wrap: break-word;
word-wrap: break-word;
```
These work on all browsers, including IE from version 5 onwards.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
https://www.w3schools.com/cssref/css3_pr_word-wrap.asp

dtaylor7: if you enter a large word or long email address on IE the format will break with this change. But we can see if testers are happy with this or not, can you reference previous JIRA ticket so testers can check that is still fixed? https://tools.hmcts.net/jira/browse/DIV-1448

PR #381, opened 2019-01-09
DIV-3469 - filter out empty values before schema validation

hosainnet:
# Description

This fixes a bug where an empty string is being accepted for required values (e.g petitioner first name/last name)

This happens when a Save and Resume occurs while on an incomplete form, where validation rules are not checked at the time
Fixes #DIV-3469

## Type of change

# Checklist:

### DIFF
--- a/app/core/fixtures/content/simple.json
+++ b/app/core/fixtures/content/simple.json
@@ -19,8 +19,10 @@
                     "hasMarriageCert": {
                         "required": "Please answer this question",
                         "invalid": "Please select an answer"
+                    },
+                    "petitionerFirstName": {
+                        "required": "Please enter first name"
                     }
-
                 }
 
             }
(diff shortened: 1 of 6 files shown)


hosainnet:
@dtaylor7 another way to fix was to do something like this

```
    "petitionerFirstName" : {
      "type" : "string",
      "pattern": ".*\\S.*"
    },
```
But that would have to be in every schema file.  But I thought since we do this already through other paths, it's better to put here to catch all future cases

dtaylor7:
@hosainnet Yes in actual fact i have techDebt to refactor some of the saveAndResume stuff which would solve this anyway, but for now LGTM

nsingamsetty:
This is tested and working as expected. Verify the ticket comments - DIV-3469 for more details. Going to merge this now.

PR #386, opened 2019-01-11
Amend petition changes - Nav to next unanswered question, fee update, exclude reason for divorce

dtaylor7:
# Description

1. [DIV-3745 - Amend Petition - direct Petitioner to next unanswered question.](https://tools.hmcts.net/jira/browse/DIV-3745)
As part of amend petition the user should be redirected to the next answered question so that they do not have answer all the questions again.

2. [DIV-3747 - Amend Petition: (4) Reason page to exclude previously used reasons.](https://tools.hmcts.net/jira/browse/DIV-3747)
As a Petitioner I want to be presented only with Facts that I can actually use on my amended petition so that I do not submit my amended application using a previously rejected reason.

3. [DIV-3749 - Amend Petition: (5) Use fee for 'Amend Petition' from fee register.](https://tools.hmcts.net/jira/browse/DIV-3749)
Petitioner is submitting an amended petition and wants to pay the appropriate fee.


app/core/helpers/steps.js
```diff
@@ -0,0 +1,38 @@
...
+    // TODO: this can be removed when all nextStep logic is moved to next function
+    nextStepCtx = yield step.interceptor(nextStepCtx, session);
+
+    // ensure step is valid
+    const [isValid] = step.validate(nextStepCtx, session);
+    if (isValid) {
+      nextStep = step.next(nextStepCtx, session);
+    }
+  } catch (error) {
+    //
```
tchow8: Maybe a log here to help debugging?

dtaylor7: no its not an "error" so to speak, just means there is no valid next question... perhaps will add a comment to explain that

app/middleware/draftPetitionStoreMiddleware.js
```diff
@@ -24,18 +28,45 @@ const options = {
...
+  const { steps } = res.locals;
+  const { session } = req;
+  const UnAnsweredStep = yield stepsHelper
+    .findNextUnAnsweredStep(steps.Start, session);
+
+  const alreadyOnCurrentStep = req.originalUrl === UnAnsweredStep.url;
+  if (alreadyOnCurrentStep) {
+    return next();
+  }
+
```
hosainnet: Info log might be useful here - just to know whether we redirected to an unanswered page or not (might be useful for support etc)

app/middleware/updateApplicationFeeMiddleware.js
```diff
@@ -44,7 +54,26 @@ const updateApplicationFeeMiddleware = (req, res, next) => {
...
+    })
+    .catch(error => {
+      logger.errorWithReq(req, 'fees_error', 'Error retrieving fees', error.message);
+      res.redirect('/generic-error');
+    });
+};
+
+
+const updateAmendFeeMiddleware = (req, res, next) => {
+  redisClient.get('commonProps.amendFee.amount')
```
hosainnet: I don't think ever set anything in redis with a key of `commonProps.amendFee.amount`, so this will always return empty

app/middleware/updateApplicationFeeMiddleware.js
```diff
@@ -44,7 +54,26 @@ const updateApplicationFeeMiddleware = (req, res, next) => {
         CONF.commonProps.applicationFee = JSON.parse(response);
         return true;
       }
-      return getFeeFromFeesAndPayments(req);
+      return getFeeFromFeesAndPayments(req, 'applicationFee');
+    })
+    .then(() => {
+      next();
```
hosainnet: Can be shortend to `.then(next)`  (and line 79)

app/steps/grounds-for-divorce/reason/index.js
```diff
@@ -87,6 +96,30 @@ module.exports = class ReasonForDivorce extends ValidationStep {
...
+          case reasons.desertion:
+            ctx.reasonForDivorceShowDesertion = false;
+            break;
+          case reasons.sep2Yr:
+            ctx.reasonForDivorceShowTwoYearsSeparation = false;
+            break;
+          case reasons.sep5Yr:
+            ctx.reasonForDivorceShowFiveYearsSeparation = false;
+            break;
+          default:
```
hosainnet: Maybe log an unknown reason?

app/middleware/updateApplicationFeeMiddleware.js
```diff
@@ -44,7 +54,26 @@ const updateApplicationFeeMiddleware = (req, res, next) => {
...
+      res.redirect('/generic-error');
+    });
+};
+
+
+const updateAmendFeeMiddleware = (req, res, next) => {
+  redisClient.get('commonProps.amendFee.amount')
+    .then(response => {
+      if (response) {
+        CONF.commonProps.amendFee = JSON.parse(response);
```
hosainnet: This is already being set in the getFeeFromFeesAndPayments in line 41 (https://github.com/hmcts/div-petitioner-frontend/pull/386/files#diff-875652841cbd89dbeb763952bc590381R41) - inside `getFeeFromFeesAndPayments`

app/core/helpers/steps.js
hosainnet: Might be useful to add info or warn just for support purposes

app/middleware/draftPetitionStoreMiddleware.js
dtaylor7: we would access log is that not enough?

app/middleware/updateApplicationFeeMiddleware.js
dtaylor7: it would not run `getFeeFromFeesAndPayments` if found in redis

app/middleware/updateApplicationFeeMiddleware.js
dtaylor7: good catch, not sure what happened to our caching here.... i have put it back

app/core/helpers/steps.js
dtaylor7: no its expected that an error will be caught here every time this code is ran, its how we find the last unanswered question...

PR #390, opened 2019-01-14
Optimise docker image build layers

slyg:
# Description

Hello, similar to the `div-respondent-frontend` project's PR 190, this PR adds the following enhancements:

- improve the ordering of the docker image build layer
- prepare for future ACR cache optimisations.

This is part of a general [pipelines optimisations initiative](https://tools.hmcts.net/confluence/display/CNP/Pipeline+optimisations) and should not functionally impact the current builds, though the docker build will be a bit longer unfortunately. However once the ACR cache optimisation is release the build time should be shortened, which is the next step in our roadmap.

Fixes [CNP-1101](https://tools.hmcts.net/jira/browse/CNP-1101)

## Type of change

- [ NA ] Bug fix (non-breaking change which fixes an issue)
- [ NA ] New feature (non-breaking change which adds functionality)
- [ NA ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ NA ] This change requires a documentation update

# How Has This Been Tested?

Local run of the generated docker runtime image.

# Checklist:

- [ NA ] My code follows the style guidelines of this project
- [ NA ] I have made corresponding changes to the documentation
- [ NA ] I have added tests that prove my fix is effective or that my feature works
- [ NA ] New and existing unit tests pass locally with my changes
- [ NA ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,10 @@ RUN addgroup --system --gid 1001 $APP_USER \
     && adduser --system --gid 1001 -uid 1001 --disabled-password --disabled-login $APP_USER \
     && chown -R $APP_USER:$APP_USER $WORKDIR \
     && apt-get update \
-    && apt-get install -y git bzip2
+    && apt-get install -y git bzip2 \
+    && yarn config set cache-folder $WORKDIR/.yarn-cache
 COPY package.json yarn.lock ./
-RUN yarn install --production \
-    && yarn cache clean
+RUN yarn install --production
 
 # ---- Build image ----
 # This images pulls the needed dev dependencies
@@ -18,10 +18,12 @@ RUN yarn install --production \
 # Removes node dependencies for next images
 # build convenience
 FROM base as build
-RUN yarn install \
-    && yarn cache clean
+RUN apt-get install -y python2.7 python-pip
+RUN yarn install && npm rebuild node-sass
 COPY . .
-RUN yarn setup && rm -rf node_modules/
+RUN yarn setup \
+    && rm -rf node_modules/ \
+    && yarn cache clean
 
 # ---- Runtime image ----
 # This is the runtime image, inheriting from
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,33 @@
-FROM node:8.12.0-stretch
+# ---- Base image ----
+FROM node:8.12.0-stretch as base
+ENV WORKDIR=/opt/app \
+    APP_USER=hmcts
+WORKDIR ${WORKDIR}
+RUN addgroup --system --gid 1001 $APP_USER \
+    && adduser --system --gid 1001 -uid 1001 --disabled-password --disabled-login $APP_USER \
+    && chown -R $APP_USER:$APP_USER $WORKDIR
+COPY package.json yarn.lock ./
+RUN yarn install --production \
+    && yarn cache clean
 
-WORKDIR /opt/app
+# ---- Build image ----
+# This images pulls the needed dev dependencies
+# then builds the different bundles and finally
+# Removes node dependencies for next images
+# build convenience
+FROM base as build
+RUN yarn install \
+    && yarn cache clean
+COPY . .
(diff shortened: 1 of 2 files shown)


slyg:
Notice that during our ACR cache use tests with @harmrai, we manage to perform docker build time in 80 seconds (for source a code modification and using this multi-staging Dockerfile).

Dockerfile
```diff
@@ -1,10 +1,37 @@
...
-
-CMD [ "yarn", "start" ]
+# ---- Build image ----
+# This images pulls the needed dev dependencies
+# then builds the different bundles and finally
+# Removes node dependencies for next images
+# build convenience
+FROM base as build
+RUN apt-get install -y python2.7 python-pip
+RUN yarn install && npm rebuild node-sass
```
hosainnet: This seems a bit redundant as we don't need dev dependencies within the image

Dockerfile
```diff
@@ -1,10 +1,37 @@
...
+# ---- Base image ----
+FROM node:8.12.0-slim as base
+ENV WORKDIR=/opt/app \
+    APP_USER=hmcts
+WORKDIR ${WORKDIR}
+RUN addgroup --system --gid 1001 $APP_USER \
+    && adduser --system --gid 1001 -uid 1001 --disabled-password --disabled-login $APP_USER \
+    && chown -R $APP_USER:$APP_USER $WORKDIR \
+    && apt-get update \
+    && apt-get install -y git bzip2 \
```
hosainnet: We we introduced this docker file initially, we tried using slim but the problem is that yarn needed git for some stuff. I know it's more ideal to use `node:8.12.0-slim` but in this case wouldn't be quicker to keep -stretch and not have this?

Dockerfile
```diff
@@ -1,10 +1,37 @@
...
+    && yarn config set cache-folder $WORKDIR/.yarn-cache
+COPY package.json yarn.lock ./
```
hosainnet: Would this base layer ever be cached given we're using `COPY` at the top here?

Dockerfile
```diff
@@ -1,10 +1,37 @@
-FROM node:8.12.0-stretch
+# ---- Base image ----
+FROM node:8.12.0-slim as base
+ENV WORKDIR=/opt/app \
+    APP_USER=hmcts
+WORKDIR ${WORKDIR}
+RUN addgroup --system --gid 1001 $APP_USER \
+    && adduser --system --gid 1001 -uid 1001 --disabled-password --disabled-login $APP_USER \
+    && chown -R $APP_USER:$APP_USER $WORKDIR \
```
hosainnet: I like the limited user stuff 👍

Dockerfile
slyg: Yes indeed, they are only used for building the project, but never end up in the runtime image

Dockerfile
slyg: So this is an optimistic instruction as it will be cached if the content of package.json and yarn.lock doesn't change. it is put on top as those files are less likely to change than any other source code file

Dockerfile
slyg: Well, once these layers are cached it won't really matter, excepted that the smaller the image, the quicker the build is. And `-slim`, with those packages, is still much smaller than `-stretch`.

Dockerfile
slyg: Now any modification of those 2 files will trigger a new yarn install. But again, it is less likely to happen than a .js/.ts/css file and avoids the npm install on each build

Dockerfile
slyg: yeah, well, we start introducing an `hmcts` user. If we end up with AKS in prod, we'll have a way to enforce policies at the cluster level more easily. Notice that all node images come with a `node` user as well :)

slyg:
All in all if this PR doesn't suit you guys (I understand it takes more time than the existing one), we can still integrate the ACR cache re-use in this same PR ok. No rush at merging it right now.

PR #394, opened 2019-01-16
Enable redirect feature

tchow8:


infrastructure/prod.tfvars
```diff
@@ -24,7 +24,7 @@ decree_nisi_frontend_url = "https://www.decree-nisi.apply-divorce.service.gov.uk
 
 capacity = "2"
 
-feature_redirect_on_state = "false"
+feature_redirect_on_state = "true"
```
qzhou-hmcts: can this flag be set to  true to all the environments?

tchow8: It'll force redirect if you have multiple cases past Submitted state.
Up to QAs

PR #399, opened 2019-01-17
Change CTSC siteId to servicecentre

tchow8:


hosainnet:
@tchow8 is it safer to put allocation to 0 for unreasonable-behaviour for the time-being?

PR #408, opened 2019-01-23
DIV-4036 improve logging

hosainnet:
# Description

Fixes #DIV-4036

## Type of change

# Checklist:


app.js
```diff
@@ -129,8 +129,7 @@ exports.init = listenForConnections => {
 
   app.use((error, req, res, next) => {
     if (error.code === 'EBADCSRFTOKEN') {
-      logger.error('csrf error has occurred', req);
-      logger.info(error);
```
gurpiarbassi: @hosainnet question on process rather than the code: - This task is a subtask of the DIV-3884 logging story. I can see that's set for fix version 5.3. Does that mean all these changes need to be being a toggle?

hosainnet: I don't think so - it's a tech debit that has no user impact so it can go in anytime. I believe feature flags are for controlling rollout of features

gurpiarbassi: fair enough.
Features and **bug fixes** too I think. I'm working on a bug fix but its targeted for 5.1. Thats the reason I asked.

hosainnet: Sure

hosainnet: I would have thought a bug fix should go in ASAP?

hosainnet: Unless of course it naturally only occures within an existing feature flag in that case it's fine

app/services/logger.test.js
```diff
@@ -95,63 +93,4 @@ describe(modulePath, () => {
       expect(loggerStub.error.calledOnce).to.eql(true);
     });
   });
-
-  describe('#idamUserIdFromReq', () => {
```
alexwakeman: Just curious why we're removing these unit tests.

app.js
gurpiarbassi: Yeh I agree.
It's a bug that's currently in live. I'll talk to Brian about it and see if he really wants it for 5.1 only.

app/services/logger.test.js
hosainnet: Removed those functions from https://github.com/hmcts/div-petitioner-frontend/pull/408/files#diff-2f876aef44739dc3746ddb1485ae379fL4

PR #413, opened 2019-01-25
Changed serviceCentre court email. This only affects the done pages

tchow8:


infrastructure/variables.tf
```diff
@@ -434,7 +434,7 @@ variable "court_service_centre_openinghours" {
 }
 
 variable "court_service_centre_email" {
-  default = "divorce@justice.gov.uk"
```
qzhou-hmcts: your probably want to update 'court_email' or delete the variable and replace it for 'court_service_centre_email', I don't know why is duplicated

tchow8: So basically, all CTSC cases should show the divorcecase email, otherwise show divorce

qzhou-hmcts: ok, then the PR looks good

PR #414, opened 2019-01-25
parseBool all featureToggles

tchow8:
parseBool all features


tchow8:
Full functional test suite passed in #415

PR #421, opened 2019-01-30
Cannot resume after answering no on screening questions

tchow8:
# Description

[DIV-4103](https://tools.hmcts.net/jira/browse/DIV-4103)

This PR removes save draft capabilities on the Screening Question pages. It also does not destroy the session on reaching a screening question exit page.

This PR also introduces the sign out functionality.

## Type of change

# How Has This Been Tested?

Local unit tests passed and manual testing. Local E2E test pass. Verifying through build pipeline.

# Checklist:


app/core/steps/DestroySessionStep.js
```diff
@@ -5,8 +5,8 @@ const CONF = require('config');
 const parseBool = require('app/core/utils/parseBool');
 
 module.exports = class DestroySessionStep extends Step {
-  * preResponse(req) {
-    yield new Promise(resolve => {
+  preResponse(req) {
+    return new Promise(resolve => {
```
dtaylor7: ?

app/steps/sign-out/index.js
```diff
@@ -0,0 +1,18 @@
...
+  get url() {
+    return '/sign-out';
+  }
+
+  get nextStep() {
+    return this.steps.Index;
+  }
+
+  handler(req, res) {
+    Promise.resolve(this.preResponse(req))
```
dtaylor7: could be:
```
yield this.preResponse(req);
res.redirect(this.next().url);
```

app/steps/sign-out/template.html
```diff
@@ -0,0 +1,17 @@
+{% extends "layout.html" %}
```
dtaylor7: presume this isnt used?

dtaylor7: this sign out page could be as easy as:
```
app.get('/sign-out', (req, res) => {
      req.session.regenerate(() => {
            res.redirect(this.next().url);
      });
});
```
in app.js?

could save the extra overhead...

tchow8: Yes, sounds good. How would I pass in the idam logout middleware here though?
I could do:
```
app.get('/sign-out', (req, res, next) => {
    req.session.regenerate(() => {
      if (parseBool(CONF.features.idam)) {
        idam.logout()(req, res, next);
      }
      res.redirect('/index');
    });
});
```
But I'm worried about the next call the idam logout middleware does will conflict with the res redirect?

tchow8: Ah nice :+1:

tchow8:
@dtaylor7 
Made some changes.
Added in a new method of the steps called stepType which returns the class name as a string. This is used by CYA so that if the nextStep is CYA or an ExitStep it won't go to it.
Moved the Sign Out to the app.js. Testing it with the E2E path.

app/steps/check-your-answers/index.js
```diff
@@ -237,12 +237,9 @@ module.exports = class CheckYourAnswers extends ValidationStep {
       //
     }
 
-    if (nextStep === this) {
+    if (nextStep === this || (nextStep && nextStep.stepType() === 'ExitStep')) {
       delete session.nextStepUrl;
-    }
-
-    // if next step and next step is not check your answers
-    if (nextStep && nextStep !== this) {
```
dtaylor7: how come you removed this. we only want to get next templates if the page is not the CYA step....

app/core/steps/ScreeningValidationStep.js
```diff
@@ -0,0 +1,27 @@
...
+
+const initSession = require('app/middleware/initSession');
+const sessionTimeout = require('app/middleware/sessionTimeout');
+const { redirectOnCondition } = require('app/middleware/redirectMiddleware');
+const { hasSubmitted } = require('app/middleware/submissionMiddleware');
+const { restoreFromDraftStore } = require('app/middleware/draftPetitionStoreMiddleware');
+const { idamProtect } = require('app/middleware/idamProtectMiddleware');
+const { setIdamUserDetails } = require('app/middleware/setIdamDetailsToSessionMiddleware');
+
+module.exports = class ScreeningValidationStep extends ValidationStep {
```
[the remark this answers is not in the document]
tchow8: They don't want the screening questions to create a draft at all now, so I did this to remove the save draft middleware from the Step.

app/steps/check-your-answers/index.js
tchow8: Thought it logically fit, the condition above checks if nextStep === this (which is CYA), so I can only run this block if nextStep is defined, since if the above condition is === this then this if block can not ever run.

app/core/steps/ScreeningValidationStep.js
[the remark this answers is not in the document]
dtaylor7: my assumption is that they requested the screen questions not to save to draft store because of the bug we are fixing now? If we are fixing it a different way i cannot see any benefit to this?

app/steps/check-your-answers/index.js
dtaylor7: fair one, the logic is correct.

Although it confused me so i guess there is an argument to make it a little simpler....

app/core/steps/ScreeningValidationStep.js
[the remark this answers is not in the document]
tchow8: I think they don't want the Screening Questions to save anyway. I could remove it from this story I suppose, but the eventual idea is the screening questions come before login to IDAM.
Although I see issues around that like, a Petitioner resuming a case has to answer the screening questions again everytime unless they use some link that goes straight to the /start route.

app/steps/check-your-answers/index.js
tchow8: Reverted the change and added the DestroySessionStep check in the template block. As you said, don't need to delete the session nextUrl I think. Will test this on PR to make sure it works though.

app/steps/check-your-answers/index.js
```diff
@@ -237,12 +237,9 @@ module.exports = class CheckYourAnswers extends ValidationStep {
       //
     }
 
-    if (nextStep === this) {
+    if (nextStep === this || (nextStep && nextStep.stepType() === 'ExitStep')) {
```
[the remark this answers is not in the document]
tchow8: Done, thanks for the advice.

app.js
```diff
@@ -145,6 +146,18 @@ exports.init = listenForConnections => {
     res.redirect('/index');
   });
 
+  // sign out route
+  const nextMiddleware = (req, res, next) => {
+    next();
+  };
+  app.get('/sign-out',
```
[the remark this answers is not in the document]
tchow8: Done

PR #423, opened 2019-01-31
Used fixed days for content test

tchow8:
Using moment.js 2 months can give a range of days between 58-61 I think.
However the test asserts for 61 days.
I've changed the subtraction to be  fixed 61 days for the failing test.


tchow8:
better fix incoming.

PR #426, opened 2019-01-31
Remove separation date line from cya with new separation dates

tchow8:
[DIV-4147](https://tools.hmcts.net/jira/browse/DIV-4147)

Removes an unnecessary line from being rendered on the CYA page


kubernetes/config.aat.yaml
```diff
@@ -60,9 +60,13 @@ data:
   COURT_SERVICE_CENTRE_EMAIL: "divorcecase@justice.gov.uk"
   COURT_SERVICE_CENTRE_PHONENUMBER: "0300 303 0642"
   COURT_SERVICE_CENTRE_SITEID: "CTSC"
-  COURT_SERVICE_CENTRE_WEIGHT: "0.30"
+  COURT_SERVICE_CENTRE_WEIGHT: "0.51"
```
alexwakeman: Curious why we're changing the weightings here?

qzhou-hmcts: I forgot to add this changes  on my PR, thanks @tchow8

tchow8: FactMap weightings changed. There is some rule where the court weighting must be equal to or greater than the fact map weightings multiplied by some values. I think @Qiang can explain more. Basically, this change will fix the AKS deployment, you'll see every PR on PFE will now fail to deploy to AKS unless this court weight change goes in.

PR #443, opened 2019-02-13
Fixed Draft Store Tests now that Screening questions do not save draft

tchow8:
Will fix failing e2e tests


tchow8:
We are only running basicDivorce on PRs based on lower resources. If AKS Preview can handle it, our only issue will be 40+ mins per PR build (unless we can find a way to speed them up), then we can probably enable all tests. Would be safer that way.

PR #444, opened 2019-02-14
Don't signout on no certificate translated exit page and updated content

tchow8:
# Description

[DIV-4108](https://tools.hmcts.net/jira/browse/DIV-4108)

This PR disables destroying the session and sign out on reaching the no translated certificate exit page.
The content on the exit page is also updated.

# Checklist:


app/steps/check-your-answers/index.js
```diff
@@ -211,7 +212,7 @@ module.exports = class CheckYourAnswers extends ValidationStep {
       delete session.nextStepUrl;
     }
 
-    if (nextStep && nextStep !== this && !(nextStep instanceof DestroySessionStep)) {
+    if (nextStep && nextStep !== this && !(nextStep instanceof DestroySessionStep) && !(nextStep instanceof ExitStep)) {
```
dtaylor7: can we break this logic out as its becoming unwieldy e.g.
```
const nextStepIsCYAStep =  nextStep == this;
const nextStepIsDestroyStep = nextStep instanceof DestroySessionStep;
const nextStepIsExitStep = nextStep instanceof ExitStep;
if (nextStep && !nextStepIsCYAStep && !nextStepIsDestroyStep && !nextStepIsExitStep) {
```

app/core/steps/ExitStep.js
```diff
@@ -0,0 +1,5 @@
+const Step = require('./Step');
+
+module.exports = class ExitStep extends Step {
```
dtaylor7: will these be confusing for future devs, the fact we have destroy session step, exit step ( which does nothing ).... page we should just extend step on the `ExitNoCertificateTranslated`

app/steps/check-your-answers/index.js
dtaylor7: dont remember logic, but presume it `ExitNoCertificateTranslated` found wont the next step be undefined anyway? whats the purpose of adding this to the if statement?

nsingamsetty:
This is tested and working fine.Will merge this PR on behalf of Phi - Please refer DIV-4108 for evidences.

app/steps/check-your-answers/index.js
tchow8: It the step before that sets the nextStep to ExitNoCertificateTranslated. So when they resume, and its not a DestroySessionStep, they resume on the Exit page, when they want to resume on the previous page.

app/core/steps/ExitStep.js
tchow8: Makes sense, I was thinking of doing that but I felt I would be duplicating code across every future exit step. But if you feel that is better, I can modify that. Or would renaming this step to like TerminalStep or something make more sense?

nsingamsetty:
This is tested and signed off. Merging this on behalf of Phi.

PR #447, opened 2019-02-18
Remove legacy error alert

tchow8:
A change to cnp-module-metric-alert broke our pipeline:
https://github.com/hmcts/cnp-module-metric-alert/commit/01cf6ff54632054d37789d8a779f2be593b599e9

However we have determined the code is no longer necessary as the issue has now been fixed, so removing the alerts to save costs and build times.


nsingamsetty:
This is the PR to fix the master build failures. Merging this now.

PR #469, opened 2019-03-06
DIV-4383 - don't allow duplicate cases in non-prod

hosainnet:
# Description

This PR removes 3 features flags

* `hasSubmittedEnabled` which allows re-submission in non-production env per user
* `redirectToApplicationSubmitted` which was only redirecting to the "application submitted" page in production if you've already submitted a petition
* `fullPaymentEventDataSubmission` which was always true anyway

This aligns non-prod envs with productions, so that the behaviour is consistent and we can replicate issues such as duplicate cases easily.

Fixes #DIV-4383

## Type of change

# Checklist:


app/middleware/submissionMiddleware.js
```diff
@@ -32,24 +30,26 @@ const handleCcdCase = (req, res, next) => {
...
     return res.redirect(APPLICATION_MULTIPLE_REJECTED_CASES_PATH);
   default:
+    logger.infoWithReq(req, 'case_done_and_submitted', 'Default case state - redirecting to done and submitted');
     return res.redirect(DONE_AND_SUBMITTED);
   }
 };
 
 const hasSubmitted = function(req, res, next) {
   const { session } = req;
-  const hasSubmittedEnabled = ['prod'].includes(config.deployment_env);
```
hosainnet: Removed prod-specific features..

app/middleware/submissionMiddleware.js
```diff
@@ -32,24 +30,26 @@ const handleCcdCase = (req, res, next) => {
...
 
   if (session.payment_reference) session.currentPaymentReference = session.payment_reference;
-  if (hasSubmittedEnabled && session.caseId && session.state) { // eslint-disable-line
+  if (session.caseId && session.state) {
     return handleCcdCase(req, res, next);
   }
   // when a new case has just been submitted for the session
-  if (parseBool(config.features.redirectToApplicationSubmitted) && session.caseId) { // eslint-line-disable
```
hosainnet: This flag was only true in production. Now, if there is a case ID, we always redirect to the application submitted page

app/services/logger.js
```diff
@@ -3,7 +3,8 @@ const { get } = require('lodash');
 
 const getIdamId = req => {
   const idamId = get(req, 'idam.userDetails.id', 'unknown');
-  return `IDAM ID: ${idamId}`;
+  const caseId = get(req, 'session.caseId', 'unknown');
+  return `IDAM ID: ${idamId}, CASE ID: ${caseId}`;
```
hosainnet: Added Case ID so we can trace the case back to a user within PFE. This will mostly be `unknown` until the petition is submitted, but will be valuable because payment logs will contain this

app/services/submission.js
```diff
@@ -52,27 +51,18 @@ const generatePaymentEventData = (session, response) => {
     { external_reference, amount, reference, status, date_created } = response; // eslint-disable-line camelcase
   // Provide status when finished, empty string otherwise.
   const siteId = get(session, `court.${session.courts}.siteId`);
-  let eventData = null;
-
-  if (parseBool(CONF.features.fullPaymentEventDataSubmission)) {
```
hosainnet: This was only true in prod..

app/services/logger.js
dtaylor7: consider wrapping in if and only adding when available

app/middleware/redirectMiddleware.test.js
```diff
@@ -27,90 +26,44 @@ describe(modulePath, () => {
...
-      featureTest(done);
+      expect(next.calledOnce).to.eql(true);
     });
 
-    it('should call next wen there is no session', done => {
-      const test = cleanup => {
-        delete req.session;
-
-        redirectMiddleware.redirectOnCondition(req, res, next);
+    it('should call next wen there is no session', () => {
```
ravinderpal: *when. Typo?

hosainnet: Ah good catch, too bad it's merged now. I guess one for a future PR :D

app/services/logger.js
hosainnet: 👍

PR #479, opened 2019-03-19
Upgrade e2e test library

vianvi:
After removal of index page on nightly pipeline webdriverIO have an issue with index page. So just removed index and upgraded library

### DIFF
--- a/test/end-to-end/paths/basicDivorce.js
+++ b/test/end-to-end/paths/basicDivorce.js
@@ -4,7 +4,7 @@ Feature('Basic divorce path');
 
 Scenario('Get a divorce', async function(I) {
 
-  I.amOnPage();
+  I.amOnPage('/');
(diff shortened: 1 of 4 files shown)


test/end-to-end/paths/basicDivorce.js
```diff
@@ -4,7 +4,7 @@ Feature('Basic divorce path');
 
 Scenario('Get a divorce', async function(I) {
 
-  I.amOnLoadedPage('/index');
+  I.amOnPage('/');
```
hosainnet: `/` will redirect to `/index` anyway so you might as well keep `/index` to save your self one redirect...

vianvi: didn't worked on webdriverIO nightly pipeline some reason, as it goes directly to idam login page, so trying with /

PR #485, opened 2019-03-22
Correct errors for marriage date page

dtaylor7:
# Description

[DIV-4500 - PFE: date fields - don't overload citizen with error messages.](https://tools.hmcts.net/jira/browse/DIV-4500)
 - Show only the correct errors at the right time
 - Fix false positive unit tests


nsingamsetty:
This is tested and working as Expected.Merging now. See ticket DIv-4500 for more details

PR #489, opened 2019-03-26
DIV-4364: Replace incorrect feature toggle in e2e tests

johnoneill46:
Fix failing e2e test that is causing master build to fail


nsingamsetty:
This is to fix master Toggle .Merging now

PR #497, opened 2019-04-01
DIV-4583 - fix incorrect step skipping on amend petition

hosainnet:
# Description

This is to fix a scenario where skipping answered questions still stops
at Check Your Answers, because the user could have answered a question,
then clicked saved an resume, which means they come back, the page is
skipped and the case is not submitted

Fixes #DIV-4583

## Type of change

# Checklist:

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,7 +76,7 @@ module.exports = function(grunt) {
         options: {
           nodeArgs: ['--trace-warnings', '--inspect'],
           ext: 'js, json, yaml',
-          ignore: ['node_modules/**', 'app/assets/**', 'public/**'],
+          ignore: ['node_modules/**', 'app/assets/**', 'public/**', 'functional-output/**', 'coverage/**', 'test/**'],
           args: grunt.option.flags()
         }
       }
--- a/app/core/helpers/steps.js
+++ b/app/core/helpers/steps.js
@@ -12,7 +12,7 @@ const getNextValidStep = function* (step, session) {
 
     // ensure step is valid
     const [isValid] = step.validate(nextStepCtx, session);
-    if (!step.isSkipWhenValid() && isValid) {
+    if (isValid) {
       nextStep = step.next(nextStepCtx, session);
     }
   } catch (error) {
@@ -27,6 +27,8 @@ const findNextUnAnsweredStep = function* (step, session = {}) {
 
   if (!nextStep) {
     return step;
+  } else if (!nextStep.isSkipWhenValid()) {
+    return nextStep;
   }
 
   return yield findNextUnAnsweredStep(nextStep, session);
--- a/app/core/helpers/steps.js
+++ b/app/core/helpers/steps.js
@@ -12,7 +12,7 @@ const getNextValidStep = function* (step, session) {
 
     // ensure step is valid
     const [isValid] = step.validate(nextStepCtx, session);
-    if (isValid) {
+    if (!step.isSkipWhenValid() && isValid) {
       nextStep = step.next(nextStepCtx, session);
     }
   } catch (error) {
--- a/app/core/helpers/steps.js
+++ b/app/core/helpers/steps.js
@@ -27,7 +27,7 @@ const findNextUnAnsweredStep = function* (step, session = {}) {
 
   if (!nextStep) {
     return step;
-  } else if (!nextStep.isSkipWhenValid()) {
+  } else if (!nextStep.shouldSkipWhenValid()) {
     return nextStep;
   }
 
--- a/app/core/helpers/steps.js
+++ b/app/core/helpers/steps.js
@@ -27,7 +27,7 @@ const findNextUnAnsweredStep = function* (step, session = {}) {
 
   if (!nextStep) {
     return step;
-  } else if (!nextStep.shouldSkipWhenValid()) {
+  } else if (!nextStep.isSkippable) {
     return nextStep;
   }
 
--- a/app/core/steps/Step.js
+++ b/app/core/steps/Step.js
@@ -89,7 +89,7 @@ module.exports = class Step {
     return [true, []];
   }
 
-  shouldSkipWhenValid() {
+  get isSkippable() {
     return true;
   }
 
--- a/app/core/steps/Step.js
+++ b/app/core/steps/Step.js
@@ -90,7 +90,7 @@ module.exports = class Step {
   }
 
   isSkipWhenValid() {
-    return false;
+    return true;
   }
 
   generateContent(ctx, session, lang = 'en') {
--- a/app/core/steps/Step.js
+++ b/app/core/steps/Step.js
@@ -89,6 +89,10 @@ module.exports = class Step {
     return [true, []];
   }
 
+  isSkipWhenValid() {
+    return false;
+  }
+
   generateContent(ctx, session, lang = 'en') {
     if (!this.content || !this.content.resources) {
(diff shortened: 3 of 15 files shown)


Gruntfile.js
```diff
@@ -76,7 +76,7 @@ module.exports = function(grunt) {
         options: {
           nodeArgs: ['--trace-warnings', '--inspect'],
           ext: 'js, json, yaml',
-          ignore: ['node_modules/**', 'app/assets/**', 'public/**'],
+          ignore: ['node_modules/**', 'app/assets/**', 'public/**', 'functional-output/**', 'coverage/**', 'test/**'],
```
hosainnet: This fixes a lot of flakiness locally because those paths constantly change, causing the app to restart and making the tests fail

app/core/helpers/steps.js
```diff
@@ -27,6 +27,8 @@ const findNextUnAnsweredStep = function* (step, session = {}) {
 
   if (!nextStep) {
     return step;
+  } else if (!nextStep.isSkipWhenValid()) {
+    return nextStep;
```
hosainnet: We don't continue finding steps if we arrive at a step which shouldn't be skipped

package.json
```diff
@@ -25,7 +25,7 @@
     "test:smoke": "./bin/run-smoke-tests.sh",
     "test:functional": "./bin/run-functional-tests.sh",
     "test:crossbrowser": "./bin/run-crossbrowser-tests.sh",
-    "test-unit": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. mocha './app/**/*.test.js' --reporter spec --recursive -t 3000",
+    "test-unit": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. mocha './app/**/*.test.js' --reporter spec --recursive -t 10000",
```
hosainnet: I found this helpped a little locally

test/end-to-end/pages/index/index.js
```diff
@@ -16,22 +16,13 @@ function startApplication(ignoreIdamToggle = false) {
...
+async function startApplicationWith(sessionName) {
 
   let I = this;
 
   I.amOnLoadedPage('/index');
   I.startApplication();
-  I.haveABasicSession('basicDivorceSessionData');
-}
-
-async function startApplicationWithAnAmendPetitionSession() {
```
hosainnet: Refactored these to be more generic

app/steps/check-your-answers/index.js
```diff
@@ -304,4 +304,8 @@ module.exports = class CheckYourAnswers extends ValidationStep {
         res.redirect('/generic-error');
       });
   }
+
+  isSkipWhenValid() {
```
dtaylor7: would never skip cya anyway, can remove this

app/core/helpers/steps.js
```diff
@@ -27,6 +27,8 @@ const findNextUnAnsweredStep = function* (step, session = {}) {
 
   if (!nextStep) {
     return step;
+  } else if (!nextStep.isSkipWhenValid()) {
```
tchow8: Maybe just me, but isSkippableWhenValid or shouldSkipWhenValid sounds better to me? haha but that is preference.

hosainnet: Yep agreed, will change it 👍

test/end-to-end/pages/index/index.js
vianvi: can we run all tests on PR?

hosainnet: Sure, just found a small issue with a unit test, I'll enable all tests on PR

app/steps/check-your-answers/index.js
hosainnet: So the scenario is, if you tick the statement of truth, then save and close (during an amend). The next time come back, it'll be "valid" so it'll skip to the pay page even though you never clicked submit (i.e there is no case)

hosainnet: Because in amend we do fast-forward all valid steps..

dtaylor7: ahh, after looking at code i remember now, yes i see what you mean

app/core/helpers/steps.js
dtaylor7: infact, may be better to put some code here: https://github.com/hmcts/div-petitioner-frontend/blob/b5c56da3096fd81bc64c85b9859f7cd94159b87a/app/core/helpers/steps.js#L2

would be more performant as wouldn't run through next step logic i.e.
```
const getNextValidStep = function* (step, session) {
  let nextStep; // eslint-disable-line init-declarations
  if (!step.isSkippable) {
    return nextStep;
  }
  ....
```

hosainnet: Correct me if I'm wrong, but I believe that would mean that whatever page we stop at wouldn't have its data pre-populated?

My initial version actually had it a few lines down the lines you've suggested (https://github.com/hmcts/div-petitioner-frontend/pull/497/commits/c619ce0a1b61d09df0b9641c216d7c7eb370d7ad#diff-abc17ad053f837e42773293384c478eaR15) however it turns out `getNextValidStep` is also used within check-your-answers page for some other logic which broke, so I thought `findNextUnAnsweredStep` is a better place because it's the only used in the amend journey (as far as I can tell) and because we _still_ want to carry out the existing validations etc..

Let me know your thoughts, happy for your input as you know this area more :)

dtaylor7: no it only populates data here to do the validation check. but if were always going to stop here then no need to validate.

test/end-to-end/codecept.conf.js
```diff
@@ -75,7 +75,7 @@ function configureChunks() {
 function getTests() {
   console.log('### CONF.preview_env =', CONF.preview_env);  // eslint-disable-line no-console
   if (CONF.preview_env === 'true') {
-    return './paths/**/basicDivorce.js';
+    return './paths/**/*.js';
```
dtaylor7: also undo this change please

hosainnet: @vianvi wanted this just to verify the build, will do once all is green!

app/core/helpers/steps.js
hosainnet: @dtaylor7 but it'll break this https://github.com/hmcts/div-petitioner-frontend/blob/master/app/steps/check-your-answers/index.js#L208

dtaylor7: yep good point, leave as it is then

app/core/steps/Step.js
```diff
@@ -89,6 +89,10 @@ module.exports = class Step {
     return [true, []];
   }
 
+  isSkipWhenValid() {
```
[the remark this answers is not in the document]
hosainnet: Done 👍

test/end-to-end/codecept.conf.js
hosainnet: Done 👍

app/core/steps/Step.js
```diff
@@ -89,6 +89,10 @@ module.exports = class Step {
     return [true, []];
   }
 
+  get isSkippable() {
```
alexwakeman: Seems all good, if it was me I would have called this `isRequired` and inverted the logic :)

dtaylor7: good point, always more readable when testing for truthy bools

PR #500, opened 2019-04-02
fix error where session is being destroyed

dtaylor7:
# Description

[DIV-4366 - Amend petition : When we move from DN to PFE - internal server error- intermittent](https://tools.hmcts.net/jira/browse/DIV-4366)
 - Fix bug where 500 error being returned to the user due to session being destroyed without being regenerated

### DIFF
--- a/app/steps/authenticated/index.js
+++ b/app/steps/authenticated/index.js
@@ -11,7 +11,7 @@ const runNext = (req, res, next) => {
   // if the previous session has expired
   // after user logged in then destroy it
   if (session.expires <= Date.now()) {
-    req.session.destroy(() => {
+    req.session.regenerate(() => {
       next();
     });
   } else {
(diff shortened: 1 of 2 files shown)


app/steps/authenticated/index.js
```diff
@@ -11,7 +11,7 @@ const runNext = (req, res, next) => {
   // if the previous session has expired
   // after user logged in then destroy it
   if (session.expires <= Date.now()) {
-    req.session.destroy(() => {
+    req.session.regenerate(() => {
```
hosainnet: Anyway to test it?

dtaylor7: set session cookie expire using browser i guess

hosainnet: Something less manual, perhaps a unit test just for regression later

dtaylor7: yes i have a unit test for this.

hosainnet: Seems a bit of a high-level test then because changing the underlying code doesn't break the test but breaks the functionality? Perhaps it should be asserting a stub

dtaylor7: 👍 have added a new assertion to ensure session has been regenerated not just deleted

PR #503, opened 2019-04-03
DIV-4615 - fix feature flags for E2E tests 

hosainnet:
# Description

1) Enable feature flags in deafult.yaml/AAT configs to always true as this is now the default for testing
2) Enable all tests to run in PRs (AKS only) to catch errors early

Fixes: #DIV-4615

## Type of change

# Checklist:


config/default.yaml
```diff
@@ -131,9 +131,9 @@ commonProps:
 features:
   idam: false
   release510: true
-  release520: false
+  release520: true
```
hosainnet: These was the main issue. When the flag is turned on in the azure portal, tests were running with release520 as false, therefore not matching the expected assertions

Jenkinsfile_CNP
```diff
@@ -24,11 +24,15 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
...
 
 withPipeline("nodejs", product, component) {
-  if (env.CHANGE_TITLE && !env.CHANGE_TITLE.startsWith('[PREVIEW]')) {
-    enableDockerBuild()
-    enableDeployToAKS()
-    loadVaultSecrets(secrets)
-    setVaultName('div')
+  if (env.CHANGE_TITLE) {
+    if(env.CHANGE_TITLE.startsWith('[PREVIEW]')) {
+      env.RUN_BASIC_TESTS = "true"
```
hosainnet: Only run basic tests in PRs tagged with `[PREVIEW]` because they're under-resourced

test/end-to-end/codecept.conf.js
```diff
@@ -4,9 +4,6 @@ const CONF = require('config');
 const waitForTimeout = parseInt(CONF.e2e.waitForTimeoutValue);
 const waitForAction = parseInt(CONF.e2e.waitForActionValue);
 
-console.log('waitForTimeout value set to', waitForTimeout); // eslint-disable-line no-console
-console.log('waitForAction value set to', waitForAction); // eslint-disable-line no-console
-
```
hosainnet: Removing noise

config/default.yaml
tchow8: I don't think this solves the issue? Its when the app settings on azure is manually changed, there is a mismatch with the infrastructure value.
I think the only solution is that app settings gets applied on every deploy, not just when the state file detects a change.

On the other hand, I do think for local development, having this set to true is better.
If we don't trust our pipeline to apply the infrastructure vars properly though, false is safer.

hosainnet: @tchow8 Hmm yeah I agree with you, false is safer in `default.yaml` - is there a way to force Azure to always apply the variables from TF files on every deploy (like K8S does)?
Maybe the other approach for manually testing feature flags is to do it on it's own branch/PR.. but that has it's issues as wel..

tchow8: not that I'm aware of I'm afraid. But if AAT is meant to be prod like anyway, we should be testing the feature on the PR itself I guess, and just verifying the feature isn't active on AAT?

hosainnet: @tchow8 so we agreed not to manually change env variables directly in Azure. I think I found a good way around feature flags locally, as we already have `development.yaml` which is used by yarn scripts. Also introduced `preview.yaml` to make sure feature flags match AKS/Preview configs - take a look again please

config/testing.yaml
```diff
@@ -1,3 +1,5 @@
+# This is used in E2E testing
```
jvictorcf: What do you mean? Master pipeline tests?

hosainnet: Master and local actually

PR #520, opened 2019-04-18
Enable strategic payment solution

tchow8:
Tested the payment service callback url with Krishna, and confirmed its not working as expected. Re-enabling the feature so we can handle future failed payment states.


tchow8:
Closing to reopen new PR as this build seems broken.

PR #526, opened 2019-04-29
Use hardcoded day value so differing month days don't throw an error

tchow8:
Basically, today's date is 29/04/2019

The failing test does:
Years - 100
Month - 1
Day - Same

However, on today's date, this results in the following date:
29/02/1919 (because month is 0 indexed, i.e. January is 00, so April is 03, so -1 results in 2, which when used in the citizen UI context relates to February)

Being a non-leap year day, 29 is not a valid date, resulting in the Invalid Date error showing instead of 100 years in the past.

This PR tries to resolve for that scenario by forcing the day component to be 1.

### DIFF
--- a/app/steps/marriage/date/index.test.js
+++ b/app/steps/marriage/date/index.test.js
@@ -13,6 +13,7 @@ const modulePath = 'app/steps/marriage/date';
 
 const content = require(`${modulePath}/content`);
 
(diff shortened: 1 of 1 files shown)


nsingamsetty:
Merging this to fix pipeline issues

PR #528, opened 2019-04-29
DIV-4499: Created new Financial Remedy page

johnoneill46:
https://tools.hmcts.net/jira/browse/DIV-4499

Created a new page in the PFE journey that informs the petitioner about settling financial matters with the respondent

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ For example:
 make test
 make test-unit
 make test-e2e
-make lint
 ...
 ```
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -44,7 +44,6 @@ For example:
 make test
 make test-unit
 make test-e2e
-make lint
```
hosainnet: Why?

johnoneill46: @hosainnet We haven't actually defined 'lint' in the makefile - it's eslint but imo I think we should get rid of the makefiles entirely. Happy to correct it if you'd prefer it to be there though

johnoneill46:
> Still needed I assume?

Yep still necessary for https://tools.hmcts.net/jira/browse/DIV-4499

PR #530, opened 2019-04-30
Enable serviceCallbackUrl again

tchow8:
Getting proper confirmation from Pay Team the exponential payment retries fix is live.

---

Confirmed updated payment code is in Production now, and current Production queue for payments is empty.

### DIFF
--- a/app/services/payment.js
+++ b/app/services/payment.js
@@ -15,8 +15,6 @@ const service = {
    */
   create: (req, user, serviceToken, caseReference, siteId, feeCode,
     feeVersion, amountInput, description, returnUrl, serviceCallbackUrl) => {
-    // Temporarily disable sending serviceCallbackUrl
-    serviceCallbackUrl = ''; // eslint-disable-line
     return client.create(user, serviceToken, caseReference, siteId, feeCode,
       feeVersion, amountInput, description, returnUrl, serviceCallbackUrl)
       .then(response => {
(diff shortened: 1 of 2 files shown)


app/services/payment.js
```diff
@@ -15,8 +15,6 @@ const service = {
    */
   create: (req, user, serviceToken, caseReference, siteId, feeCode,
     feeVersion, amountInput, description, returnUrl, serviceCallbackUrl) => {
-    // Temporarily disable sending serviceCallbackUrl
-    serviceCallbackUrl = ''; // eslint-disable-line
```
hosainnet: I think we should have a unit test for this time around (if not integration)..

tchow8: Unit test won't catch anything since its a REST call. I could write one but it would be pretty pointless like expect(create).to.be.calledWith(mockedServiceCallbackUrl);
It would already be tested as part of the payment e2e flow, since the issue is on Payment side, and there is no response to PFE from the serviceCallback.
I feel adding tests for this change would just be unnecessarily adding time taken to the test suite.

hosainnet: The fact that no test was breaking by overriding the URL to `''` seems like we're missing coverage though.. I feel like having a test which asserts that the all params are passed onto `client.create` is a valid one

tchow8: Updated the unit test to check parameters as well.

hosainnet: Thanks 👍

PR #535, opened 2019-05-07
Feature Toggle Strategic Callback

tchow8:
Issue where payment team is immediately triggering the serviceCallbackUrl when receiving payments, without waiting for the scheduler to kick in.

Disabling the serviceCallbackUrl until they resolve.

### DIFF
--- a/app/services/payment.js
+++ b/app/services/payment.js
@@ -15,6 +15,8 @@ const service = {
    */
   create: (req, user, serviceToken, caseReference, siteId, feeCode,
     feeVersion, amountInput, description, returnUrl, serviceCallbackUrl) => {
+    // temporarily disable strategic payment callback whilst bug is being resolved.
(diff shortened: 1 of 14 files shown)


qzhou-hmcts:
Probably we need a feature flag for this...

hosainnet:
@qzhou-hmcts what is the point given that our flow requires a code change anyway? If we used feature flags properly (like change config in prod and restart on the fly) then I'd agree with you

qzhou-hmcts:
@hosainnet  that can still be done, I'm quite sure when ccd sees 20~30 reqs/second devops  will be  happy to update the feature flag and restart the service. 
Having the feature flag you can give a full context on the functionality  that is not given by 'serviceCallbackUrl'

tchow8:
Reviews again please. Only 1 request every few mins now though :D

PR #536, opened 2019-05-07
Statement of truth field should never be checked

mounikahmcts:
https://tools.hmcts.net/jira/browse/DIV-4800

PFE-CYA page - Statement of Truth check box is always checked for particular scenario which is wrong


nsingamsetty:
This is tested and signed off.  Refer DIV-4800 for more details and evidences. Merging this code

PR #541, opened 2019-05-15
DIV-3532 fix missing divorce options

hosainnet:
# Description

- Renamed `start-dev` to `dev` to match DN and RFE
- Changed port to 3000 to match DN and RFE 
- Run the interceptor as part of Save and Close to return any calulcated values there into the session

Fixes #DIV-3532

## Type of change

# Checklist:

### DIFF
--- a/app/middleware/draftPetitionStoreMiddleware.js
+++ b/app/middleware/draftPetitionStoreMiddleware.js
@@ -196,26 +196,28 @@ const saveSessionToDraftStoreAndClose = function(req, res, next) {
   const hasSaveAndCloseBody = body && body.saveAndClose;
 
   if (isPost && hasSaveAndCloseBody) {
-    let ctx = this.parseRequest(req); // eslint-disable-line no-invalid-this
-    ctx = this.interceptor(ctx); // eslint-disable-line no-invalid-this
-    const session = this.applyCtxToSession(ctx, req.session); // eslint-disable-line no-invalid-this
-    const sessionToSave = removeBlackListedPropertiesFromSession(session);
+    co(function* generator() {
+      let ctx = this.parseRequest(req); // eslint-disable-line no-invalid-this
+      ctx = yield this.interceptor(ctx); // eslint-disable-line no-invalid-this
+      const session = this.applyCtxToSession(ctx, req.session); // eslint-disable-line no-invalid-this
+      const sessionToSave = removeBlackListedPropertiesFromSession(session);
 
-    // Get user token.
-    let authToken = '';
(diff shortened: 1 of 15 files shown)


app/middleware/draftPetitionStoreMiddleware.js
```diff
@@ -196,25 +196,30 @@ const saveSessionToDraftStoreAndClose = function(req, res, next) {
...
-      })
-      .catch(error => {
-        logger.errorWithReq(req, 'save_draft_and_close_error', 'Error restoring draft', error.message);
-        res.redirect('/generic-error');
-      });
+    const context = this; // eslint-disable-line no-invalid-this
+
+    co(function* generator() {
+      let ctx = context.parseRequest(req);
+      ctx = yield context.interceptor(ctx);
```
alexwakeman: just trying to wrap my head around this 😝 - so the data which the user was expecting to save wasn't being saved to session? So we have to invoke the steps interceptor to get values because the POST `req` doesn't contain the values in the right format..? Is that right?

nsingamsetty:
This is tested and signed off. Ready to Merge.

app/middleware/draftPetitionStoreMiddleware.js
```diff
@@ -196,25 +196,30 @@ const saveSessionToDraftStoreAndClose = function(req, res, next) {
...
+      // Get user token.
+      let authToken = '';
+      if (req.cookies && req.cookies[authTokenString]) {
+        authToken = req.cookies[authTokenString];
+      }
+
+      const sendEmail = true;
+      client.saveToDraftStore(authToken, sessionToSave, sendEmail)
+        .then(() => {
+          res.redirect(context.steps.ExitApplicationSaved.url); // eslint-disable-line no-invalid-this
```
qzhou-hmcts: minor:probably the comment is not needed anymore

app/middleware/draftPetitionStoreMiddleware.js
hosainnet: As far as I understand, any context data calculated on the fly in the interceptor (i.e not directly taken from user input, e.g `divorceWho` or `marriageDate`) was not being set on the context (and therefore not copied to the session) because we were not calling the interceptor as part of the save & close route

alexwakeman: Cool, makes sense thanks!

nsingamsetty:
This is tested and working as Expected. Refer DIV-3532 for sign off and evidences.Merging now.

PR #544, opened 2019-05-20
Update feeToBePaid if it is not available in session

mounikahmcts:
https://tools.hmcts.net/jira/browse/DIV-4743

In PFE saving /pay/online page and then relogin  will not retain feeToBePaid in session.

### DIFF
--- a/app/steps/pay/pay-online-only/index.test.js
+++ b/app/steps/pay/pay-online-only/index.test.js
@@ -62,7 +62,8 @@ describe(modulePath, () => {
 
   describe('success', () => {
     it('renders the content from the content file', done => {
-      testContent(done, agent, underTest, content);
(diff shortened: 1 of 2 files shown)


app/steps/pay/pay-online-only/index.test.js
```diff
@@ -62,7 +62,8 @@ describe(modulePath, () => {
 
   describe('success', () => {
     it('renders the content from the content file', done => {
-      testContent(done, agent, underTest, content);
+      const dataContent = { feeToBePaid: '550' };
+      testContent(done, agent, underTest, content, dataContent);
```
hosainnet: Is it possible to add a test for when `previousCaseId` is present to check that `amendFee` is used instead please?

PR #550, opened 2019-05-28
DIV-4622 screen reader reads case reference ID as seperate characters

CoffeeOwl97:
# Description

Previously when a screen reader would read the case reference Id it would ready numeric characters as numbers, e.g. an Id of `TOM123` would be read as `TOM one hundred and twenty three`, instead of `T O M one two three`. This change adds a nunjucks filter which spaces out each of the characters which is applied to an ARIA label, so that when a screen reader tries to read it out, the version that the screen reader reads will have a whitespace inbetween each character, causing it to read out each character indevidualy.

Fixes # (issue)
[Accessibility [AA] - Case ID Number](https://tools.hmcts.net/jira/browse/DIV-4622)

## Type of change

# How Has This Been Tested?

Enable a screen reader and check output when on the case identification number.
Unit testing of new filter functionality.

# Checklist:


app/filters/nunjucks.test.js
```diff
@@ -11,4 +11,17 @@ describe('Nunjucks insert hyphen filter', () => {
     const filteredString = nunjucksFilters.insertHyphens(caseIdNumber);
     expect(filteredString).to.equal('1509 &#8208; 0317 &#8208; 9378 &#8208; 0148');
   });
-});
\ No newline at end of file
+});
+
+describe('a11yCharSeparator', () => {
+  it('should return string formatted', () => {
```
yarpo: what about empty string, number and null?

app/filters/nunjucks.js
```diff
@@ -5,6 +5,18 @@ const newFilterInsertHyphen = input => {
   return `${input}`.match(/.{1,4}/g).join(' &#8208; ');
 };
 
-const newobjectfilter = { insertHyphens: newFilterInsertHyphen };
 
-module.exports = newobjectfilter;
\ No newline at end of file
+const a11yCharSeparator = input => {
+  if (typeof input === 'string') {
+    // Regex replace to remove all excess whitespace (e.g. '  a      b ' -> 'ab')
```
yarpo: why not like this:

```
const removeWhitespacesRegex = /\s+/g;
const textWithoutWhitespaces = input.replace(removeWhitespacesRegex);

return returnString.split('').join(' ')
```

Isn't it clear enough now? Without comments? :)

app/filters/nunjucks.test.js
```diff
@@ -11,4 +12,32 @@ describe('Nunjucks insert hyphen filter', () => {
...
+  });
+  it('should handle a null value', () => {
+    const filteredString = nunjucksFilters.a11yCharSeparator(null);
+    const expectedResult = null;
+    expect(filteredString).to.eql(expectedResult);
+  });
+  it('should handle a number value', () => {
+    const filteredString = nunjucksFilters.a11yCharSeparator(1234);
+    const expectedResult = 1234;
+    expect(filteredString).to.eql(expectedResult);
```
yarpo: shouldn't expected result be '1234'? (string, not int)

CoffeeOwl97: Test is perhaps a little deceptively named, the a11yCharSeparator is designed to seperate characters in a string. If the input is not strictly equal to a string (in this case a number), then it should just return that input as the output.

nsingamsetty:
Tested and working fine.Merging now.

PR #563, opened 2019-07-01
DIV-5103 - Update email address for divorce cases

dtaylor7:
# Description

[DIV-5103 - Update email address for divorce cases](https://tools.hmcts.net/jira/browse/DIV-5103)
 - change contactdivorce@justice.gov.uk to divorcecase@justice.gov.uk

### DIFF
--- a/charts/div-pfe/values.yaml
+++ b/charts/div-pfe/values.yaml
@@ -23,7 +23,7 @@ nodejs:
         COURT_SERVICE_CENTRE_CITY: "Nottingham"
         COURT_SERVICE_CENTRE_POBOX: "PO Box 10447"
         COURT_SERVICE_CENTRE_POSTCODE: "NG2 9QN"
-        COURT_SERVICE_CENTRE_EMAIL: "contactdivorce@justice.gov.uk"
+        COURT_SERVICE_CENTRE_EMAIL: "divorcecase@justice.gov.uk"
         COURT_SERVICE_CENTRE_OPENINGHOURS: "Telephone Enquiries from: 8.30am to 5pm"
         COURT_SERVICE_CENTRE_PHONENUMBER: "0300 303 0642"
         COURT_SERVICE_CENTRE_SITEID: "CTSC"
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -394,7 +394,7 @@ variable "court_service_centre_openinghours" {
 }
 
 variable "court_service_centre_email" {
-  default = "contactdivorce@justice.gov.uk"
+  default = "divorcecase@justice.gov.uk"
 }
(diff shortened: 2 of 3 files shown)


infrastructure/variables.tf
```diff
@@ -394,7 +394,7 @@ variable "court_service_centre_openinghours" {
 }
 
 variable "court_service_centre_email" {
-  default = "contactdivorce@justice.gov.uk"
+  default = "divorcecase@justice.gov.uk"
```
qzhou-hmcts: You need to update aks config to see this see  changes on preview environment .
charts/div-pfe/values.yaml

dtaylor7: my understanding was defaults get set in case orch service?

qzhou-hmcts: variables.tf only applies to terraform, however aks chart config does not go through terraform

dtaylor7: have updated, but if case orch service sets it wont make a difference

charts/div-pfe/values.yaml
```diff
@@ -23,7 +23,7 @@ nodejs:
         COURT_SERVICE_CENTRE_CITY: "Nottingham"
         COURT_SERVICE_CENTRE_POBOX: "PO Box 10447"
         COURT_SERVICE_CENTRE_POSTCODE: "NG2 9QN"
-        COURT_SERVICE_CENTRE_EMAIL: "contactdivorce@justice.gov.uk"
```
yarpo: What about line 19:
```
        COURT_EMAIL: "contactdivorce@justice.gov.uk"
```

dtaylor7: only use the new email once case is submitted

infrastructure/variables.tf
```diff
@@ -394,7 +394,7 @@ variable "court_service_centre_openinghours" {
 }
 
 variable "court_service_centre_email" {
```
yarpo: what about line 416:
```
variable "court_email" {
  default = "contactdivorce@justice.gov.uk"
}
```

dtaylor7: same as https://github.com/hmcts/div-petitioner-frontend/pull/563#discussion_r299089941

charts/div-pfe/values.yaml
yarpo: oh, ok. Sorry. My mistake then

dtaylor7:
@yarpo 
> Can you please also update:
> div-petitioner-frontend/test/end-to-end/data/amendPetitionSession.js:122
Already done

> div-petitioner-frontend/bin/run-functional-tests.sh
this needs to stay the same as its used before the case is created

PR #579, opened 2019-07-29
DIV-4001 - Upgrade to the latest nodejs version

dtaylor7:
# Description

[DIV-4001 - Upgrade to the latest nodejs version](https://tools.hmcts.net/jira/browse/DIV-4001)
 - The latest version installed on our infrastructure is: 10.15.2


package.json
```diff
@@ -81,7 +81,7 @@
     "js-yaml": "^3.6.1",
     "jsonwebtoken": "^8.2.2",
     "loadsh": "^0.0.3",
-    "lodash": "^4.17.13",
+    "lodash": "^4.17.5",
```
yarpo: why?

dtaylor7: why not?

`yarn upgrade` i presume

Dockerfile
```diff
@@ -1,7 +1,16 @@
...
 USER root
 RUN apt-get update && apt-get install -y bzip2 git python2.7 python-pip
-COPY package.json yarn.lock ./
+COPY package.json yarn.lock .nvmrc ./
+
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash \
+    && chmod +x $HOME/.nvm/nvm.sh \
+    && . $HOME/.nvm/nvm.sh
+
+RUN nvm install \
```
timja: should really use the node provided in the base image, and nvmrc should just match that
https://github.com/hmcts/cnp-node-base

Use `hmctspublic.azurecr.io/base/node/stretch-slim-lts-10:10-stretch-slim` as the base

package.json
timja: I assume @yarpo is saying why downgrade it

Dockerfile
dtaylor7: that would be amazing if it work, been struggling with this all day 👍

package.json
dtaylor7: Yep, for some reason i read it as 1 3 rather than 13. I have reverted must of happened when i was reviewing resolutions 👍

PR #584, opened 2019-08-02
DIV-5406: Fixing bug from "DIV-4238: PFE - use court information from COS response"

jvictorcf:
Reverts hmcts/div-petitioner-frontend#575

**So, here's the story...**

I've pushed hmcts/div-petitioner-frontend/pull/573 to prod, but we've got prod error when petitioners were logging back into the app after finishing the divorce application.

So, to get production bug-free ASAP, we've reverted this PR with this PR: hmcts/div-petitioner-frontend/pull/575 (thanks @qzhou-hmcts)
We then raised a bug to investigate what went wrong with the original PR. And this is what this current PR is about.

**What went wrong?**

I didn't realise that there is one bit of the PFE app that actually cares about how the court data is represented in the COS query response. As it turns out, redirectMiddleware.js does.
So, to make sure we cater either the new format (coming from the /submit response) or the old format (coming from the COS query response), I've made sure we always check both in the redirectMiddleware.js.

Please, feel free to leave comments and feedback.


nsingamsetty:
This is tested and working fine. Refer the evidences for ticket DIV-5406.Merging now.

PR #604, opened 2019-08-28
Fix chart syntax for IDAM tagging

alexwakeman:
[Fix IDAM chart tags syntax to allow for post-PR scripts](https://tools.hmcts.net/jira/browse/DIV-5694)


alexwakeman:
Closed as it causes issues in IDAM whitelisting

PR #608, opened 2019-09-06
DIV-5719 - duplicate payment fix

hosainnet:
# Description

Scnearios:

1) Just before redirecting to the payment website - we get all payment attempts for a case - if we cannot get the payment history - we display an error page

2) Just before redirecting to the payment website - we get all payment attempts for a case - there is a success payment already (but we couldn't update the case the firs time) - we skip the payment page and update the case that a payment was made

3) Just before redirecting to the payment website - we get all payment attempts for a case - we find initiated payments (i.e either they didn't yet complete the details OR they did but we were not able to verify that for whatever reason) - we display the following new placeholder page (content needs verification)

 
Screenshot 2019-09-12 at 11 37 41

Because we cannot differentiate between the 2 possibilites in scenario 3 (either payment not completed or completed but not verified), it's safer to display this page above than allow them to pay again (which is what is happening now).

But this means that if they stop half way through the payment journey (i.e they don't enter card details and close the page), and they attempt to login again within ~90 minutes and complete the payment, we will not allow them to proceed and display the above page (after 90 minutes the payment API marks "initiated" payment sessions to failed).

I've attached the 2 videos of the user flow before/after the fix in the Jira story, hope that makes more sense

Fixes #DIV-5719 

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

Unit tests, also running locally and mocking the Payment API service (see video attached to DIV-5719 and the steps to reproduce there)

# Checklist:


package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "NODE_PATH=. node server.js",
     "start-basic": "NODE_ENV=production NODE_PATH=. node server",
-    "dev": "NODE_ENV=development NODE_PATH=. grunt dev",
+    "dev-aat": "http_proxy=http://proxyout.reform.hmcts.net:8080 https_proxy=http://proxyout.reform.hmcts.net:8080 NODE_ENV=aat NODE_PATH=. grunt dev",
```
tchow8: Do we need to remove the existing dev script? Is this what the earlier comment is about?

nsingamsetty:
This is tested and evidences can be seen on the ticket DIV-5719. Merging now.

PR #617, opened 2019-09-13
PROD FIX: parse feature as bool so it copes with strings

dtaylor7:
# Description

Fix prod issue where webchat feature is being passed as a string

### DIFF
--- a/app.js
+++ b/app.js
@@ -27,6 +27,7 @@ const logging = require('app/services/logger');
 const events = require('events');
 const idam = require('app/services/idam');
 const signOutRoute = require('app/routes/sign-out');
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -111,7 +112,7 @@ exports.init = listenForConnections => {
     loader: nunjucks.FileSystemLoader,
     globals: {
       webchat: CONF.services.webchat,
```
charlesluokb: what about this? no need to use parseBool?

dtaylor7: yes we need to parse bool as on prod the value is "0" which in JS is parsed as true

dtaylor7: this isnt for feature but to provide client with service keys for webchat

PR #622, opened 2019-10-01
DIV-5761 Add test and extra checks for invalid API data

alexwakeman:
[Invalid address information breaks petition](https://tools.hmcts.net/jira/browse/DIV-5761)


alexwakeman:
This change has been regression tested in PR

app/components/AddressLookupStep/addressTypes/postcode.js
```diff
@@ -96,7 +96,7 @@ module.exports = {
 
       if (ctx.selectAddressIndex !== '-1' && ctx.addresses && ctx.addresses.length) {
         const address = ctx.addresses[ctx.selectAddressIndex];
-        if (address && address.DPA.ADDRESS) {
+        if (address && address.DPA && address.DPA.ADDRESS) {
```
hosainnet: What is the effect visually on the page if address.DPA is not present? i.e do they not get the address populated or do they have to enter it manually?

alexwakeman: We see this if the data is malformed -treated as an error state. Thanks for the review. 

image

app/components/AddressLookupStep/addressTypes/postcode.js
```diff
@@ -73,7 +73,7 @@ module.exports = {
         addresses,
         error
       } = yield postcodeLookupClient.lookupPostcode(ctx.postcode);
-      if (error) {
+      if (error || (Array.isArray(addresses) && addresses.length > 0 && !addresses[0].DPA)) {
```
dtaylor7: move this condition to a variable which explains what it is!

alexwakeman: ok cool done @dtaylor7

nsingamsetty:
This is tested and working fine. Merging now.

PR #624, opened 2019-10-07
Prepare for AKS migration

timja:


charts/div-pfe/values.yaml
```diff
@@ -1,20 +1,52 @@
...
+        GOOGLE_ANALYTICS_ID: "TBD" # override for prod TBD in all other envs
+        PAYMENT_REFERENCE_SERVICE_IDENTIFICATION: "DIV1"
+        PAYMENT_SERVICE_HEALTHCHECK_URL: "http://payment-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/health"
+        PAYMENT_SERVICE_URL: "http://payment-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
+        REFORM_ENVIRONMENT: "{{ .Values.global.environment }}"
+        SERVICE_AUTH_PROVIDER_HEALTHCHECK_URL: "http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/health"
+        SERVICE_AUTH_PROVIDER_URL: "http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
+        IDAM_APP_HEALHCHECK_URL: "https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net/health"
+        IDAM_API_URL: "https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net"
```
hosainnet: This won't work for prod, will it?

timja: no  it'll need overriding

charts/div-pfe/values.preview.template.yaml
```diff
@@ -9,33 +10,19 @@ idam-pr:
...
-        EVIDENCE_MANAGEMENT_CLIENT_API_URL: "http://div-emca-aat.service.core-compute-aat.internal"
         FEATURE_IDAM: "true"
         FEATURE_STRATEGIC_PAY: "false"
-        FEES_AND_PAYMENTS_HEALTHCHECK_URL: "http://div-fps-aat.service.core-compute-aat.internal/health"
-        FEES_AND_PAYMENTS_URL: "http://div-fps-aat.service.core-compute-aat.internal"
-        GOOGLE_ANALYTICS_ID: "TBD"
-        PAYMENT_REFERENCE_SERVICE_IDENTIFICATION: "DIV1"
-        PAYMENT_SERVICE_HEALTHCHECK_URL: "http://payment-api-aat.service.core-compute-aat.internal/health"
-        PAYMENT_SERVICE_URL: "http://payment-api-aat.service.core-compute-aat.internal"
         REDISCLOUD_URL: "redis://${SERVICE_NAME}-redis-master:6379"
```
hosainnet: This can go

charts/div-pfe/values.preview.template.yaml
```diff
@@ -9,33 +10,19 @@ idam-pr:
...
     environment:
         APPINSIGHTS_INSTRUMENTATIONKEY: "dummykey"
-        CASE_ORCHESTRATION_SERVICE_DRAFT_URL: "http://div-cos-aat.service.core-compute-aat.internal/draftsapi/version/1"
-        CASE_ORCHESTRATION_SERVICE_HEALTHCHECK_URL: "http://div-cos-aat.service.core-compute-aat.internal/health"
-        CASE_ORCHESTRATION_SERVICE_URL: "http://div-cos-aat.service.core-compute-aat.internal"
-        DECREE_NISI_FRONTEND_URL: "https://div-dn-aat.service.core-compute-aat.internal"
-        DEPLOYMENT_ENV: "preprod"
-        EVIDENCE_MANAGEMENT_CLIENT_API_HEALTHCHECK_URL: "http://div-emca-aat.service.core-compute-aat.internal/health"
-        EVIDENCE_MANAGEMENT_CLIENT_API_URL: "http://div-emca-aat.service.core-compute-aat.internal"
         FEATURE_IDAM: "true"
```
hosainnet: This can go as it's the default

charts/div-pfe/values.preview.template.yaml
```diff
@@ -9,33 +10,19 @@ idam-pr:
...
-        REFORM_ENVIRONMENT: "aat"
-        SERVICE_AUTH_PROVIDER_HEALTHCHECK_URL: "http://rpe-service-auth-provider-aat.service.core-compute-aat.internal/health"
-        SERVICE_AUTH_PROVIDER_URL: "http://rpe-service-auth-provider-aat.service.core-compute-aat.internal"
-        IDAM_APP_HEALHCHECK_URL: "https://idam-api.aat.platform.hmcts.net/health"
-        IDAM_API_URL: "https://idam-api.aat.platform.hmcts.net"
-        IDAM_AUTHENTICATION_HEALHCHECK_URL: "https://idam-web-public.aat.platform.hmcts.net/health"
-        IDAM_LOGIN_URL: "https://idam-web-public.aat.platform.hmcts.net/login"
-        # TODO - this should be changed once we deploy AAT to AKS
-        BASE_URL: "https://${SERVICE_NAME}.service.core-compute-preview.internal"
         PUBLIC_HOSTNAME: "https://${SERVICE_NAME}.service.core-compute-preview.internal"
```
hosainnet: This can go

charts/div-pfe/values.yaml
```diff
@@ -1,20 +1,52 @@
...
+        IDAM_AUTHENTICATION_HEALHCHECK_URL: "https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/health"
+        IDAM_LOGIN_URL: "https://idam-web-public.{{ .Values.global.environment }}.platform.hmcts.net/login"
+        PUBLIC_HOSTNAME: "https://div-pfe-{{ .Values.global.environment }}.service.core-compute-preview.internal" # override in public facing envs
```
hosainnet: `-preview`

charts/div-pfe/values.yaml
```diff
@@ -1,20 +1,52 @@
 nodejs:
     applicationPort: 3000
+    ingressHost: "div-pfe-{{ .Values.global.environment }}.service.core-compute-preview.internal" # override in public facing environments
```
hosainnet: `-preview`

charts/div-pfe/values.preview.template.yaml
timja: isn't it needed on preview?

charts/div-pfe/values.yaml
timja: don't you want it set to your specific PR? if so it needs to stay in the preview one

timja: ah i see

charts/div-pfe/values.preview.template.yaml
hosainnet: It's defaulted in values.yaml I think

charts/div-pfe/values.preview.template.yaml
timja: can it? don't you need it on your PR?

charts/div-pfe/values.preview.template.yaml
timja: i removed that, values.yaml isn't remplated

charts/div-pfe/values.preview.template.yaml
timja: i.e this will have pr-624-div-pe-preview...

hosainnet: Interesting, I don't know why we would use this this format

hosainnet: But I probably wouldn't want to touch it now..

app/services/healthcheck.js
```diff
@@ -40,22 +40,22 @@ const setup = app => {
             logger.errorWithReq(null, 'Health check failed on redis:', error);
           });
       }),
-      'idam-api': healthcheck.web(config.services.idamApp.health,
+      'idam-api': healthcheck.web(`${config.idamArgs.idamApiUrl}/health`,
```
timja: will this have any impact on your team @hosainnet ?
I think there's a stub which may not have /health on each app by looking at the local config, but the problem is you aren't using the actual app config that way.

I misconfigured it in AKS and the app config was right but the health check was failing, bad design imo. This PR doesn't remove all the config but it just only stops looking at it

timja: discussed on slack, Hosain said it should be fine

PR #635, opened 2019-10-28
exclude AddressLookupStep/mocks/postcodeInfo.js from sonar checks

yarpo:
# Description

Fix sonar coverage check by excluding mock file.


yarpo:
it's properly fixed by PR-637

PR #637, opened 2019-10-29
Fix code cov for AddressLookupStep

alexwakeman:
Adds extra tests to ensure code cov for newer code.

### DIFF
--- a/app/components/AddressLookupStep/index.js
+++ b/app/components/AddressLookupStep/index.js
@@ -29,7 +29,11 @@ module.exports = class AddressLookupStep extends ValidationStep {
   }
 
   hasSessionSelectedAddress(session) {
-    const hasPostcodeLookup = session.postcodeLookup && session.postcodeLookup.addresses && session.postcodeLookup.selectAddressIndex;
+    const hasPostcodeLookup = !([
+      session.postcodeLookup,
+      session.postcodeLookup.addresses,
+      session.postcodeLookup.selectAddressIndex
+    ].includes(undefined)); // eslint-disable-line no-undefined
     if (hasPostcodeLookup) {
       return session.postcodeLookup.addresses[session.postcodeLookup.selectAddressIndex] && session.postcodeLookup.addresses[session.postcodeLookup.selectAddressIndex].DPA;
     }
--- a/app/components/AddressLookupStep/index.js
+++ b/app/components/AddressLookupStep/index.js
@@ -29,23 +29,24 @@ module.exports = class AddressLookupStep extends ValidationStep {
   }
 
   hasSessionSelectedAddress(session) {
+    const currentSession = session[this.schemaScope];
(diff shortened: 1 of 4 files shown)


app/components/AddressLookupStep/index.js
```diff
@@ -29,19 +29,24 @@ module.exports = class AddressLookupStep extends ValidationStep {
   }
 
   hasSessionSelectedAddress(session) {
-    const hasPostcodeLookup = session.postcodeLookup && session.postcodeLookup.addresses && session.postcodeLookup.selectAddressIndex;
+    const currentSession = session[this.schemaScope];
+    const hasPostcodeLookup = !([
```
jvictorcf: Granted, I'm not a JS guy, but wouldn't this read more easily if it was a private method?
If this is a common JS way of asserting this, I'm ok with that.

sonar-project.properties
```diff
@@ -5,7 +5,7 @@ sonar.projectVersion=0.0.1
 sonar.sources=app
 sonar.tests=test
 sonar.language=js
-sonar.exclusions=app/assets/**, app/**/*.test.js, test/**, **/healthcheck.js, app/steps/save-resume/confirm-remove-saved-application/index.js, app/middleware/draftPetitionStoreMiddleware.js, app/steps/sitemap/client.js, app/steps/error-400/template.html, app/steps/error-404/template.html, app/views/question.html, app/steps/marriage/upload/template.html, app/steps/check-your-answers/template.html, app/steps/done-and-submitted/template.html
+sonar.exclusions=app/assets/**, app/**/*.test.js, test/**, **/healthcheck.js, app/steps/save-resume/confirm-remove-saved-application/index.js, app/middleware/draftPetitionStoreMiddleware.js, app/steps/sitemap/client.js, app/steps/error-400/template.html, app/steps/error-404/template.html, app/views/question.html, app/steps/marriage/upload/template.html, app/steps/check-your-answers/template.html, app/steps/done-and-submitted/template.html, app/components/AddressLookupStep/mocks/postcodeInfo.js
```
jvictorcf: Why are we excluding this file?

app/components/AddressLookupStep/index.js
alexwakeman: this is a known way of asserting all values are or are not with a certain value - we need this becuase `currentSession.selectAddressIndex` is often `0` which is `falsey` in JS - so standard truthy test won't work

https://stackoverflow.com/a/43635855/239036 reference.

sonar-project.properties
alexwakeman: it's only a test support file, not app code - we don't need coverage testing on it.

app/components/AddressLookupStep/index.js
jvictorcf: Got you. Thanks.

PR #657, opened 2020-02-03
RPET-47: Removing pages and questions about respondent's solicitor.

jvictorcf:
# Description

RPET-47: Removing pages and questions about respondent's solicitor.

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Unit tests

# Checklist:


app/steps/respondent/correspondence/send-to-solicitor/partials/checkYourAnswersTemplate.html
```diff
@@ -1,9 +1,7 @@
 <tr class="govuk-table__row">
   <th class="govuk-table__cell" scope="row">{{ content.question | safe }}</th>
   <td class="govuk-table__cell">
-    {% if fields.respondentSolicitorRepresented.value === 'Yes' %}
-      {{ content.solicitor | safe }}
-    {% elseif fields.respondentSolicitorRepresented.value === 'No' %}
+    {% if fields.respondentSolicitorRepresented.value === 'No' %}
```
johnoneill46: Whats the rational of keeping this IF statement? Can it not be removed entirely?

jakub-kozlowski: it can still make sense if `fields.respondentSolicitorRepresented.value` might be empty - if it's always 'No' then could be removed.

jvictorcf: That's a fair comment. Maybe we shouldn't even show this only mandatory option to the user (since he has no choice but to choose it). However, I'm assuming that we don't want to rethink the user journey as part of this ticket.
In this case, because I'm still presenting this only choice to the user, I think I'd rather keep the view counting on that choice. In the practical sense, you are, of course, right.

What do you think?

yarpo: Seems to be a bit redundant, but it's not critical. If we know that this field is always populated here with 'No' then it can be remove. Are we sure it's always true?

PR #741, opened 2020-07-23
Div-6155: PFE awaiting amend landing page

Dizrupto:
# Description

Adding landing page for when case is in: AwaitingAmendCase

Fixes https://tools.hmcts.net/jira/browse/DIV-6155

## Type of change

# Checklist:


app/services/documentHandler.js
```diff
@@ -0,0 +1,18 @@
+const config = require('config');
+const idam = require('app/services/idam');
+const { initDocumentHandler } = require('@hmcts/div-document-express-handler');
+
+const documentWhiteList = config.document.documentWhiteList;
+
+const initDocumentHandlerFor = app => {
+  const middleware = [ idam.protect() ];
+  const args = {
+    documentServiceUrl: `${config.evidenceManagmentClient.url}${config.evidenceManagmentClient.downloadEndpoint}`,
```
jvictorcf, line 10: typo: evidenceManagmentClient

app/services/documentHandler.js
```diff
@@ -0,0 +1,18 @@
+const config = require('config');
+const idam = require('app/services/idam');
+const { initDocumentHandler } = require('@hmcts/div-document-express-handler');
+
+const documentWhiteList = config.document.documentWhiteList;
+
+const initDocumentHandlerFor = app => {
```
jvictorcf, line 7: What does this do?

app/middleware/redirectMiddleware.test.js
```diff
@@ -49,7 +49,15 @@ describe(modulePath, () => {
...
-    it('should call redirect to DN if the state is not AwaitingPayment', () => {
+    it('should call next when the state is AwaitingAmendCase', () => {
+      req.session.state = 'AwaitingAmendCase';
+
+      redirectMiddleware.redirectOnCondition(req, res, next);
+
+      expect(next.calledOnce).to.eql(true);
+    });
+
+    it('should call redirect to DN if the state is not AwaitingPayment or AwaitingAmendCase', () => {
```
jvictorcf, line 60: I thought we said that cases in AwaitingAmendCase should stay in PFE and show the explanatory page.
Shouldn't it only be forwarded to DN when the case is actually being amended?

jvictorcf: It might just be the test's description that's wrong.

app/steps/authenticated/index.js
```diff
@@ -37,9 +38,14 @@ module.exports = class Authenticated extends Step {
   }
 
   nextStep(session) {
+    if (isAwaitingAmendCase(session)) {
```
jvictorcf, line 41: Didn't we say that this was too early to ask this question?
The session is only loaded in the steps below, isn't it?

app/steps/awaiting-amend/content.json
```diff
@@ -0,0 +1,69 @@
...
+          "decreeNisiLink": "<a class=\"govuk-link\" href=\"https://www.gov.uk/divorce/apply-for-decree-nisi\" target=\"_blank\" aria-label=\"Decree nisi link, This link will open in a new tab.\">Decree nisi</a>",
+          "decreeAbsoluteLink": "<a class=\"govuk-link\" href=\"https://www.gov.uk/divorce/apply-for-a-decree-absolute\" target=\"_blank\" aria-label=\"Decree absolute link, This link will open in a new tab.\">Decree absolute</a>",
+          "childrenAndDivorceLink": "<a class=\"govuk-link\" href=\"https://www.gov.uk/looking-after-children-divorce/overview\" target=\"_blank\" aria-label=\"Children and divorce link, This link will open in a new tab.\">Children and divorce</a>",
+          "moneyAndPropertyLink": "<a class=\"govuk-link\" href=\"https://www.gov.uk/money-property-when-relationship-ends\" target=\"_blank\" aria-label=\"Money and property link, This link will open in a new tab.\">Money and property</a>"
+        },
+        "errors": {}
+      }
+    },
+    "cy": {
```
jvictorcf, line 36: Are we letting someone know about this? Someone has to translate it, right?

app/steps/awaiting-amend/template.html
```diff
@@ -0,0 +1,120 @@
...
+</p>
+
+{{ hmctsProgressBar({
+items: [
+{
+label: { text: content.youApply },
+complete: true
+},
+{
+label: { text: content.yourRespondes },
```
jvictorcf, line 41: typo

app/steps/awaiting-amend/template.html
```diff
@@ -0,0 +1,120 @@
+{% from "warning-text/macro.njk" import govukWarningText %}
```
jvictorcf, line 1: I think I understand. You've moved the explanation page to be the view of this step. Nice!
If that's right, I would not call this step "awaiting-amend". I'd call it something more meaningful, like "amendment-explanatory-page".

jvictorcf: I still think I wouldn't do it this way. This is not really a validation page, is it?
I won't oppose to it, but it feels a bit out of place to me.

app/steps/index/index.js
```diff
@@ -9,6 +10,10 @@ module.exports = class Index extends Step {
   }
 
   nextStep(session) {
+    if (isAwaitingAmendCase(session)) {
```
jvictorcf, line 13: why do you need this here if the authenticate page redirects you there?

app/views/common/components/awaitingAmendsComponent.html
```diff
@@ -0,0 +1,31 @@
+{#
```
jvictorcf, line 1: Can we rename this file to something more meaningful?

jvictorcf: Is this the bit showing the documents to download?

config/default.yaml
```diff
@@ -164,6 +165,7 @@ apps:
 ccd:
   d8States:
     - 'AwaitingPayment'
+    - 'AwaitingAmendCase'
```
jvictorcf, line 168: ditto

app/steps/index/index.js
madjava: @jvictorcf If this is not done it would fall through

madjava: @jvictorcf If user has already been authenticate e.g. hits a reload/refresh button after logging, the authenticate process would be bypassed. So we need this in both places.

app/services/documentHandler.js
Dizrupto: This one is responsible for initialising the library that will display the document (i.e. here the d8petition which the user can see from the explanatory page).

app/middleware/redirectMiddleware.test.js
Dizrupto: Yep that's right. Maybe I can describe the test as "if the state is not AwaitingPayment or not AwaitingAmendCase"

app/steps/authenticated/index.js
madjava: At this point the session already exists with data, this would have been done by some functionality in PFE before we get to this part

app/steps/index/index.js
jvictorcf: I believe it will only load the session because you've made the page into a ValidationStep. If it's put in the next middleware (where it actually already is), it shouldn't fail, should it?

app/steps/awaiting-amend/content.json
madjava: Yes, Jeremy is informed comments in the ticket also. I not 100% sure who but yes it would be.

app/views/common/components/awaitingAmendsComponent.html
madjava: The file contains two html components widget used in the page template. Other files in this space bare similar naming and thought to make this obvious that the components in this file are for the awaiting amends layout hence the name. Maybe and 's' at the end would be convey more meaning.

app/steps/index/index.js
madjava: Well, Marc and i want to prove this. If so we'd remove this logic here. From my experience though and also having conversation with Luca and Tony Chow, this was the best place to catch this. Would update once we have conclusive fact on this.

madjava: Ok, seems we can live without it here.

app/steps/awaiting-amend/template.html
Dizrupto: It is just the name of the text style. It makes it bold with an exclamation mark, which is what we are looking for according to the story.

jvictorcf: Oh no. I've written the comment on this line, but I'm actually talking about this file (not this line)

app/views/common/components/awaitingAmendsComponent.html
jvictorcf: Cool. If you can rename it, that'd be great.

config/default.yaml
madjava: Yup, @Dizrupto I think this can be reverted

app/steps/awaiting-amend/template.html
Dizrupto: Oow okay I see yep. I will rename the file that should be okay.
I think Felix used ValidationStep because from this page a request will be made to amend the petition.
The 'ValidationStep' has the validate() function that tests I think whether the request is valid.
A normal 'Step' would not have that.
I can see that a page like 'check-your-answers' for example also uses 'ValidationStep'. I guess probably for the same reasons, that it sends a request to a backend service.

config/default.yaml
Dizrupto: I think this one needs to be here because the redirectMiddleWare is called as part of the 'ValidationStep' middlewares. If it's not there I think there is a risk it would redirect it to DN unless the court id is null. But I can see cases where the court would be set though the case is in 'awaiting amend'.
image

app/views/common/components/awaitingAmendsComponent.html
Dizrupto: So I add the 's' at the end here?

config/default.yaml
madjava: Sorry @Dizrupto in my comment above meant to say this cant be reverted, you just pointed why. Apologies.

app/middleware/submissionMiddleware.js
```diff
@@ -35,6 +37,13 @@ const handleCcdCase = (req, res, next) => {
   case 'MultipleRejectedCases':
     logger.infoWithReq(req, 'multiple_cases_rejected', 'Multiple cases rejected');
     return res.redirect(APPLICATION_MULTIPLE_REJECTED_CASES_PATH);
+  case 'AwaitingAmendCase':
+    if (isToggleOnAwaitingAmend(session)) {
+      logger.infoWithReq(req, 'awaiting_amend_case', 'Awaiting amend case');
+      return res.redirect(AMENDMENT_EXPLANATORY_PAGE);
+    }
+    logger.infoWithReq(req, 'case_done_and_submitted', 'Default case state - redirecting to done and submitted');
```
jvictorcf, line 45: This state should not exist until the feature is switched on. Thinking about it, do you really need a feature-toggle in PFE?

app/views/common/components/amendmentExplanatoryPageComponents.html
```diff
@@ -0,0 +1,31 @@
+{#
```
jvictorcf, line 1: Cool. Thanks.
Just a suggestion: what about something like "explanatoryPageDownloadableFiles.html"?

app/services/evidenceManagmentService.js
```diff
@@ -5,7 +5,7 @@ const logger = require('app/services/logger').logger(__filename);
 const errors = require('app/resources/errors');
```
jvictorcf, line 5: Can you please rename this file and its test. There's a typo.

app/steps/amendment-explanatory-page/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+const BASE_PATH = '/';
+
+module.exports = class AwaitingAmend extends ValidationStep {
+  get url() {
+    return '/amendment-explanatory-page';
+  }
+
+  get middleware() {
+    return [
```
jvictorcf, line 19: Why do you need these middlewares? Do you not need to also use the parent class' middlewares (i.e. super.middleware())?

app/middleware/submissionMiddleware.js
Dizrupto: Technically it is unlikely indeed that a case would be in that state until release. Was it a business requirement that we have the feature toggle I guess?

app/views/common/components/amendmentExplanatoryPageComponents.html
Dizrupto: Well the thing is that's not only what it does :D.
It displays downloadablefiles, but the other macro is the 'continue' button

jvictorcf: Got you. Makes sense.

app/middleware/submissionMiddleware.js
madjava: One of the requirements was to feature toggle this. Was more pressure to get this sorted especially that time when issue was burning across HMCTS. Jeremy/Izzy would be expecting a toggle for this and i guess it would have to be turned on on PFE also, i have left a comment in the ticket :)

jvictorcf: I don't think so. We need to feature-toggle a feature, not necessarily every piece of code that delivers it.
By having more feature-toggles than we need, we're making our life release when release time comes.

If you think we need this, sure, let's keep it, but I don't think we do.
Having a case in this state before release is not only unlikely, it's impossible, isn't it?

jvictorcf: Happy to have @yarpo's take on this.

app/steps/amendment-explanatory-page/index.js
Dizrupto: That's one for @madjava :D

madjava: It would run these middleware, if not overridden then the parent ones, the parent one has functionality we don't want to run e.g. the **hasSubmitted** and **saveSessionToDraftStoreAndClose**.
If it was not overridden here then parent one would run, in some other pages thats what was expected.
-Best of my experience and knowledge answer-

jvictorcf: I see. Running the **hasSubmitted** would probably cause and infinite loop, right?

jvictorcf: **saveSessionToDraftStoreAndClose** should be alright I think as it depends on a user input...

jvictorcf: I understand your reasons for overwriting the middleware in your page. However, as I understand, it seems the whole point of extending ValidationStep is to get those middlewares for free, isn't it?
On the other hand, I understand that calling those middlewares would probably leave you in an infinite loop (because you'll be constantly redirected to the new page.

If the above is right, I think your page should not extend ValidationStep, but Step. This way, you can pick which middlewares you want to run.

Do bear in mind that, by the time the user is redirected to your page, a few of them will have already run. Here's my take on the middlewares you've specified:
idamProtect: will have already run, but it's probably ok to keep to keep away malicious users
checkCookiesAllowed: not sure why you need it. it's already used in pages before this
initSession: ditto
sessionTimeout: ditto
setIdamUserDetails: ditto

Hence, you probably just need your class to extend Step and have only the idamProtect middleware.

app/services/evidenceManagmentService.js
jvictorcf: Great

app/steps/amendment-explanatory-page/index.js
jvictorcf: As per our conversation, I'll approve as long as this is handled in a different sub-task.

app/middleware/submissionMiddleware.js
jvictorcf: As per our conversation, I'll approve as long as this is handled in a different sub-task.

PR #745, opened 2020-07-28
Div 6193: Amend case from amendment explanatory page

Dizrupto:
# Description

When clicking the 'continue' button on the amendment explanatory page developed in https://tools.hmcts.net/jira/browse/DIV-6155, the case should be amended, and the petitioner should be able to update details for their petition on a new (i.e. amended) case.

Fixes https://tools.hmcts.net/jira/browse/DIV-6193

## Type of change

# Checklist:


app/services/submission.test.js
```diff
@@ -239,4 +248,60 @@ describe(modulePath, () => {
...
+          done();
+        })
+        .catch(error => {
+          done(error);
+        });
+    });
+
+    it('rejects if something goes wrong', done => {
+      // Arrange.
+      const error = new Error('Something went wrong');
```
jacmurphy, line 292: Are we able to get any more info out of the error? It seems a bit generic

app/services/submission.test.js
```diff
@@ -239,4 +248,60 @@ describe(modulePath, () => {
...
+      sinon.stub(mockedClient, 'amend').resolves(amendSuccess);
+      client = underTest.setup();
+    });
+
+    afterEach(() => {
+      mockedClient.amend.restore();
+    });
+
+    it('forwards the call to the client', done => {
+      // Act.
```
jvictorcf, line 265: What's this comment about?

app/services/submission.test.js
```diff
@@ -239,4 +248,60 @@ describe(modulePath, () => {
...
+        .catch(error => {
+          done(error);
+        });
+    });
+
+    it('resolves sending required data of the response', done => {
+      // Act.
+      client.amend()
+        .then(output => {
+          // Assert.
```
jvictorcf, line 281: Maybe we don't need those?

app/services/submission.test.js
```diff
@@ -28,6 +28,15 @@ describe(modulePath, () => {
       status: 'success'
     }
   };
+  const amendSuccess = {
+    status: 200,
+    statusCode: 200,
```
jvictorcf, line 33: do we need both status and statusCode?

app/steps/amendment-explanatory-page/index.test.js
```diff
@@ -154,27 +157,64 @@ describe(modulePath, () => {
...
-        expires: Date.now()
+  describe('#postRequest', () => {
+    let req = {};
+    let res = {};
+    beforeEach(() => {
+      req = {
+        body: {},
+        method: 'POST',
+        session: { featureToggles: {}, state: 'AwaitingAmendCase', caseId: '123' },
+        cookies: { '__auth-token': 'eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbiI6IjEyMzQ1Njc4OTAiLCJpZCI6IjEwIn0.sX3u4V8vPF6brlIfF6-k5JxZNI_Yjz__kfHnG9MyOu4' },
```
yarpo, line 168: do we need to pass so complex token? 

```
'__auth-token': 'fake.token' },
```

wouldn't be enough? :)

[don't change it. It's fine :) ]

app/steps/amendment-explanatory-page/index.js
```diff
@@ -58,12 +60,23 @@ module.exports = class AwaitingAmend extends ValidationStep {
...
 
-  submitApplication() {
-    return true;
+  submitApplication(req, res) {
+    const authToken = req.cookies['__auth-token'];
+    const submission = submissionService.setup();
+    submission.amend(req, authToken, req.session.caseId)
+      .then(response => {
+        logger.infoWithReq(req, 'amendment_success', 'Case amended successfully', response);
+        const unAnsweredStep = stepsHelper.findNextUnAnsweredStep(this, req.session);
```
jvictorcf, line 74: unanswered

app/steps/amendment-explanatory-page/index.test.js
```diff
@@ -154,27 +157,64 @@ describe(modulePath, () => {
...
+        yield underTest.postRequest(req, res);
+        expect(underTest.parseCtx.calledOnce).to.equal(true);
+        expect(underTest.validate.calledOnce).to.equal(true);
+        expect(underTest.submitApplication.calledOnce).to.equal(true);
+        done();
+      });
+    });
+
+    it('does not submit application if invalid', done => {
+      co(function* generator() {
```
jvictorcf, line 208: what's this co function?

madjava: It's a warper function that take a generator and handles it like a promise. Still don't know fully why they decided to used them as its used in many places, but this much i know they use it to deal with the async nature of NodeJS, everything within the block would is guaranteed to run to completion.
-best of my knowledge and experience answer-

app/services/submission.test.js
Dizrupto: This one doesn't really matter it's just the test one. The log in the code is different

app/services/submission.test.js
Dizrupto: Not sure, that one was just a copy and paste from the previous way it was used, not sure why they have both.

app/middleware/submissionMiddleware.js
```diff
@@ -42,8 +43,7 @@ const handleCcdCase = (req, res, next) => {
       logger.infoWithReq(req, 'awaiting_amend_case', 'Awaiting amend case');
       return res.redirect(AMENDMENT_EXPLANATORY_PAGE);
     }
-    logger.infoWithReq(req, 'case_done_and_submitted', 'Default case state - redirecting to done and submitted');
-    return res.redirect(DONE_AND_SUBMITTED);
+    return res.redirect(CONTACT_DIVORCE_TEAM_PATH);
```
jvictorcf, line 46: What's the reason for this redirect? Just out of curiosity.

app/steps/amendment-explanatory-page/index.js
```diff
@@ -58,12 +63,39 @@ module.exports = class AwaitingAmend extends ValidationStep {
...
-      return this.submitApplication();
+      return this.submitApplication(req, res);
     }
     return yield super.postRequest(req, res);
   }
 
-  submitApplication() {
-    return true;
+  submitApplication(req, res) {
+    const authToken = req.cookies['__auth-token'];
```
jvictorcf, line 72: Isn't there a function to return this?

app/steps/amendment-explanatory-page/index.js
```diff
@@ -58,12 +63,39 @@ module.exports = class AwaitingAmend extends ValidationStep {
...
+    const submission = submissionService.setup();
+
+    return submission.amend(req, authToken, req.session.caseId)
+      .then(response => {
+        logger.infoWithReq(req, 'amendment_success', 'Case amended successfully', response);
+
+        const retainedProps = this.retainedAfterNewSessionCreated(req);
+
+        req.session.regenerate(() => {
```
jvictorcf, line 81: why do you have to regenerate the session? wouldn't it be enough to add the response to the existing session?
Should the new state really be null? Why?

app/steps/amendment-explanatory-page/index.js
```diff
@@ -58,12 +63,39 @@ module.exports = class AwaitingAmend extends ValidationStep {
...
+          redirectToNextPage(req, res);
+        });
+      })
+      .catch(error => {
+        logger.errorWithReq(req, 'amendment_error', 'Error during amendment step', error.message);
+        res.redirect('/generic-error');
+      });
+  }
+
+  retainedAfterNewSessionCreated(req) {
```
jvictorcf, line 92: If you do keep this, I'd name it as a verb. For example: getRetainedProps...().

test/fixtures/mockAmendedSession.json
```diff
@@ -0,0 +1,135 @@
+{
```
jvictorcf, line 1: Should this have a state?

app/steps/amendment-explanatory-page/index.js
madjava: You, sounds fine. We would keep it, changed in the other PR

test/fixtures/mockAmendedSession.json
madjava: It should not as COS does not return one after the amend had been done the. Key info is the previousCaseId

app/steps/amendment-explanatory-page/index.js
madjava: When we do that that way Express marks the session as updated "touched" the lots of "bad things" begin to happen. Only way was to regenerate it then add the new case data from COS in. Also looping through old session data to pick out only case data was not viable, PFE uses the session data for both holding case data and also session related functionality. Best/advised solution by Tony was this. Added some more comments in the other PR

app/middleware/submissionMiddleware.js
madjava: If case is awaiting amend but toggle is turned off, the user would still not get to the new screen they would be redirected to submitted page by default. My thinking was to go the contact page. I guess we can have a conversation about this is ok, then can be reverter @yarpo @Dizrupto @jvictorcf Please comment

app/steps/amendment-explanatory-page/index.js
Dizrupto: Can't seem to find one yet. Other places seem to take it directly like this also.

jvictorcf: Fair enough.

app/middleware/submissionMiddleware.js
jvictorcf: I see.
To be honest, as per our last conversation, I don't even think the feature-toggle applies here. However, if you want to deal with this in the next story (the clean-up one), I'm fine with it.

jvictorcf: ...As it will never happen. You won't even be able to demo this (except by mocking data)

PR #769, opened 2020-11-03
Chore: e2e and crossbrowser version bumps

Rob-Doherty:
# Description

This change updates CodeceptJS, Puppeteer and Webdriver to latest or nearly latest stable versions.

As part of this, updates were made to 
- follow CodeceptJS's move to `Webdriver` helper
- fix ElementExist to wait for element
- update crossbrowser tests to run `@nightly` tagged tests instead of hardcoded to _basicDovorce.js_
- fix saucelab.conf.js's waitForTimeout reference
- update supportedBrowsers.js to test on the latest Mac and Windows OSs

Fixes # https://tools.hmcts.net/jira/browse/TA-155

## Type of change

# How Has This Been Tested?

This has been run locally against AAT and via Jenkins build, with both passing.

**Test Configuration**:

* Hardware:
* O/S and version:
* JDK:

# Checklist:

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -165,8 +165,8 @@ saucelabs:
   username: 'username'
   key: 'privatekey'
   tunnelId: 'reformtunnel'
-  waitForTimeout: 60000
-  smartWait: 5000
+  waitForTimeout: 20000
+  smartWait: 20000
 
 public:
   protocol: 'https'
--- a/test/end-to-end/paths/reasonsForDivorce.js
+++ b/test/end-to-end/paths/reasonsForDivorce.js
@@ -116,7 +116,7 @@ languages.forEach( language => {
 
   }).retry(2);
 
-  xScenario(`${language.toUpperCase()} - 5 years separation E2E @nightly `, async function(I) {
+  Scenario(`${language.toUpperCase()} - 5 years separation E2E @nightly `, async function(I) {
 
     const divorceReason = language === 'en' ? contentEn : contentCy;
     await loginPageToEnterAddressUsingPostcode(I, language);
--- a/test/end-to-end/paths/reasonsForDivorce.js
+++ b/test/end-to-end/paths/reasonsForDivorce.js
@@ -121,7 +121,7 @@ languages.forEach( language => {
     const divorceReason = language === 'en' ? contentEn : contentCy;
     await loginPageToEnterAddressUsingPostcode(I, language);
     I.selectReasonForDivorce(language, divorceReason['5YearsSeparationHeading']);
-    I.enterSeparationDateNew(fiveYearsAgoFormatted.day, fiveYearsAgoFormatted.month, fiveYearsAgoFormatted.year,
+    I.enterSeparationDateNew(language, fiveYearsAgoFormatted.day, fiveYearsAgoFormatted.month, fiveYearsAgoFormatted.year,
       fiveYearsAgoFormatted.day, fiveYearsAgoFormatted.month, fiveYearsAgoFormatted.year);
     I.selectLivingApartTime(language);
     I.enterLegalProceedings(language);
--- a/test/end-to-end/saucelabs.conf.js
+++ b/test/end-to-end/saucelabs.conf.js
@@ -27,7 +27,7 @@ const getBrowserConfig = (browserGroup) => {
 };
 
 const setupConfig = {
-  tests: './paths/**/basicDivorce.js',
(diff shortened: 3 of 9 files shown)


mounikahmcts (approved):

phaniperla (approved):
Subject to PR green build

test/end-to-end/saucelabs.conf.js
```diff
@@ -26,17 +27,17 @@ const getBrowserConfig = (browserGroup) => {
...
-    WebDriverIO: {
+    WebDriver: {
       url: process.env.E2E_FRONTEND_URL || CONF.e2e.frontendUrl,
       browser,
       waitForTimeout,
       smartWait,
       cssSelectorsEnabled: 'true',
       host: 'ondemand.eu-central-1.saucelabs.com',
-      port: 80,
+      // port: 80,
```
Rob-Doherty, line 40: Just fyi: I've left this commented as it's something I need to address when merging the next PR in after this one, so it will get cleaned up after this is merged in

config/default.yaml
```diff
@@ -165,8 +165,8 @@ saucelabs:
   username: 'username'
   key: 'privatekey'
   tunnelId: 'reformtunnel'
-  waitForTimeout: 60000
-  smartWait: 5000
+  waitForTimeout: 20000
+  smartWait: 20000
```
mounikahmcts, line 169: is it really required to wait for extra 20 sec?

mounikahmcts (dismissed):
looks good to me 👍

test/end-to-end/saucelabs.conf.js
```diff
@@ -26,17 +27,17 @@ const getBrowserConfig = (browserGroup) => {
 };
 
 const setupConfig = {
-  tests: './paths/**/basicDivorce.js',
+  tests: './paths/**/*.js',
```
phaniperla, line 30: you need to add below code snippet then it will run both english and welsh tests 

tests: './paths/**/reasonsForDivorce.js',

phaniperla (commented):
you need to tag '@nigtly' to the feature

test/end-to-end/paths/reasonsForDivorce.js
```diff
@@ -116,7 +116,7 @@ languages.forEach( language => {
 
   }).retry(2);
 
-  xScenario(`${language.toUpperCase()} - 5 years separation E2E @nightly `, async function(I) {
+  Scenario(`${language.toUpperCase()} - 5 years separation E2E @nightly `, async function(I) {
```
phaniperla, line 119: you need to tag ' @nighly' to the feature

phaniperla: see below code snippet :Feature(' Reasons for divorce E2E Tests @functional  @nightly').retry(3);

phaniperla (approved):
Except two small changes , everything it looks good to me

config/default.yaml
Rob-Doherty: This is for the nightly tests only. Unfortunately we can have some sever delays overnight between our end, saucelabs and the network between so the timeout needs to be big enough. But having a timeout of 60secs is very slow to get feedback, so I was thinking 20 seconds is a happy balance.

test/end-to-end/saucelabs.conf.js
Rob-Doherty: The reason I've set the `tests` value to be so open-ended is that it then allows these tests to be extended pretty simply in future by just adding the `@nightly` tag to other test names.
The package.json command has been updated to `--grep` for that tag and ONLY run those tests.

test/end-to-end/paths/reasonsForDivorce.js
Rob-Doherty: At the moment there's only 1 test running nightly as per the original `basicDivorce.js` setup, but with the addition of running in both English and Welsh. If we want to add more tests to the nightly build, I propose that's done in a separate ticket after all these fixes and PRs are merged in to sort out the framework. Adding tests later should then be more straightforward

mounikahmcts (approved):

test/end-to-end/paths/reasonsForDivorce.js
phaniperla: Agree with you @Rob-Doherty

phaniperla (approved):
Approved

PR #836, opened 2021-03-15
I have no idea why, but works for me

yarpo:
Don't merge. Just to see what to do.

To run locally in UI (including debug mode):

Screenshot 2021-03-15 at 18 17 00

### DIFF
--- a/test/accessibility/a11yTest.js
+++ b/test/accessibility/a11yTest.js
@@ -53,51 +53,51 @@ for(let i in languages) {
       (function(step) {
         let results;
 
-        describe(`GET Requests - Verify accessibility for the page ${step.name} - ${languages[i]}`, () => {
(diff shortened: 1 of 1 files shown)


test/accessibility/a11yTest.js
```diff
@@ -53,51 +53,51 @@ for(let i in languages) {
...
+        //       .filter((warn) =>
+        //         !step.ignorePa11yWarnings.includes(warn.code)
+        //       );
+        //
+        //     expect(warnings.length).to.equal(0, JSON.stringify(warnings, null, 2));
+        //   });
+        //
+        // });
+
+        if (step instanceof ValidationStep && step.name === 'RespondentSolicitorDetails') {
```
yarpo, line 100: it will execute only the step that was failing

PR #838, opened 2021-03-16
Moved some functional tests to nightly build stage

madjava:
# Description

Marked some functional test on PFE to run nightly. This should improve Jenkins build and also reduce build time and number or IDAM users our test creates on each build run.

# Checklist:


test/end-to-end/paths/aboutYourMarriageCertificate.js
```diff
@@ -38,7 +38,7 @@ Scenario('Marriage certificate not in English, answered No', (I) => {
   I.seeInCurrentUrl('/exit/about-your-marriage/no-certificate-translated');
 });
 
-Scenario('@overnight: Married in UK, not answered', (I) => {
+Scenario('Married in UK, not answered @nightly', (I) => {
```
jvictorcf, line 41: I don't think this should be in nightly build. It's a core part of the app.

test/end-to-end/paths/logout.js
```diff
@@ -2,7 +2,7 @@ const CONF = require('config');
 const idamConfigHelper = require('test/end-to-end/helpers/idamConfigHelper.js');
 const parseBool = require('app/core/utils/parseBool');
 
-Feature('Logout Session @functional').retry(3);
+Feature('Logout Session @nightly').retry(3);
```
jvictorcf, line 5: Maybe this shouldn't be in nightly either.

package.json
```diff
@@ -29,7 +29,6 @@
     "test-unit": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. mocha './app/**/*.test.js' --reporter spec --recursive -t 15000 --exit",
     "test-validation": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. mocha test/validation/ --reporter spec --recursive --timeout 15000 --exit",
     "test-e2e": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. ./node_modules/.bin/codeceptjs run-multiple parallel -c test/end-to-end/codecept.conf.js --grep @functional --steps --reporter mocha-multi",
-    "test-overnight-e2e": "NODE_ENV=testing LOG_LEVEL=OFF NODE_PATH=. codeceptjs run -c test/end-to-end/codecept.conf.js --steps",
```
yarpo, line 32: we don't need it. It's not used anywhere.

test/end-to-end/paths/aboutYourMarriageCertificate.js
yarpo, line 41: ```@nightly```

is the only correct way.

jvictorcf (commented):
In general, I don't think this is the appropriate solution.
We should make creating users more explicit, in my opinion.
Also, we should only create new users when needed.

I don't really like pushing things to nightly tests, but for some of these, I'd have less of an issue.

test/end-to-end/paths/aboutYourMarriageCertificate.js
yarpo: Please see yarn command:
```
test-crossbrowser-e2e
```

https://github.com/hmcts/div-petitioner-frontend/pull/838/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R32

in Jenkinsfile_nightly we call:
```
withNightlyPipeline("nodejs", product, component) {
    env.TEST_URL = params.URL_TO_TEST

    enableCrossBrowserTest()
    enableMutationTest()

    after('crossBrowserTest') {
        steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'functional-output/**/*'
    }
}
```

```    enableCrossBrowserTest()```

and you can see in Jenkins log:
```
22:10:20  + yarn test:crossbrowser
```

log from: https://build.platform.hmcts.net/view/DIV/job/HMCTS_Nightly_DIV/job/div-petitioner-frontend/job/master/359/consoleFull

yarpo (approved):

test/end-to-end/paths/aboutYourMarriageCertificate.js
yarpo: But when somebody added it it was already meant to be executed overnight.

As all Welsh tests (please mind The Welsh team didn't add ANY e2e tests, and Test Automation team was meant to add them to nightly build - that was suggested by them and agreed. I don't know why they eventually added all their tests to  "normal" functional).

test/end-to-end/paths/logout.js
yarpo: If only you are able to unblock us in given circumstances I'm happy to run even more functional test if only they don't block us. We have 2 sprints and too slow build. We must be able to get a green build. 

What is more this is minor feature from perspective of Divorce and the risk we could break it is minimal.

Logout is not even our feature (it's idam). This should be spotted and fixed by Idam team.

test/end-to-end/paths/aboutYourMarriageCertificate.js
jvictorcf: Fair enough.

madjava: @jvictorcf Is this worth running every time there is a commit or at least one a day to make sure the functionality is not broken.

test/end-to-end/paths/logout.js
jvictorcf: We can unblock ourselves if we're given time to work on this.
It's a matter of priorities.

That doesn't mean it's right to move these to nightly build, in my opinion.

test/end-to-end/paths/aboutYourMarriageCertificate.js
jvictorcf: Every time, I'd say.

test/end-to-end/paths/logout.js
madjava: Also this @jvictorcf, its functionality that we most likely would touch in any one ticket, maybe running once a day to make sure all is ok is sufficient.

test/end-to-end/paths/aboutYourMarriageCertificate.js
jvictorcf: But I accept you're not changing the behaviour, only fixing the fact the "overnight" doesn't work.
So I'm ok with that (being pragmatic).

test/end-to-end/paths/logout.js
jvictorcf: I don't think so. We wouldn't make that decision if it wasn't for performance reasons.

madjava: Well maybe not, if this was being written now maybe we would have decided to run it at night. The performance issues just made us rethink some existing solutions

Dizrupto (approved):

PR #950, opened 2022-01-27
DFR-779 Confidential address fix

anatolysan:
https://tools.hmcts.net/jira/browse/DFR-779


arun-hmcts (approved):

arun-hmcts (commented):
@anatolysan Is there a way we can avoid setting up the  respondentContactDetailsConfidential in the session instead of removing it ... Just a thought ..I am happy to approve it

ruban72 (approved):

PR #1019, opened 2022-05-23
DFR-1086 Check labels and run tests

BWeeks-JusticeGovUK:
Check github labels and run appropriate tests

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -1,5 +1,7 @@
 #!groovy
 
+import uk.gov.hmcts.contino.GithubAPI
+
 @Library("Infrastructure")
 
 String product = "div"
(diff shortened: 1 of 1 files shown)


Jenkinsfile_CNP
```diff
@@ -30,6 +32,45 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
   ]
 }
 
+def labels = [
+  "enable_mutation_test",
```
arun-hmcts, line 36: Mutation test is not required

BWeeks-JusticeGovUK: It's entirely optional.  If the mutation test is not required for a given PR, don't add the label and it won't be executed.

BWeeks-JusticeGovUK: Removed

